PR #31, opened 2020-02-20
Replace container-interop

weierophinney:
This patch updates the package to act as a replacement for container-interop, ~~by adding `class_alias` statements via an autoloader that alias the various container-interop interfaces to the psr/container equivalents~~ by importing the last release of container-interop/container-interop directly into the package.

This allows us to remove the functionality within the abstract plugin manager that decorates the parent container as a container-interop-based container if it is a psr/container, ~~and essentially allows creates a dual-inheritance tree that allows either interface to be used at any time, regardless of the typehint~~ as any container passed as a parent container now is guaranteed to fulfill either container-interop or PSR-11, and the plugin manager only consumes the PSR-11 interface internally.

~~I've tested it within another application, and had no errors result. I think this may be a BC way to eliminate our usage of container-interop, but would like others to test as well.~~

~~If it works, I can also look into updating all other typehints used within the package, and validate that extensions work as expected.~~

I have added my repo as a Composer repo, specifying my branch as "3.4.4"; when installed in this way, I no longer get warnings about installing container-interop, and all code continues to work.

Finally, I have also updated the command for generating factories to no longer generate factories implementing `Laminas\ServiceManager\Factory\FactoryInterface`. This allows them to typehint against PSR-11 for the container, making factories generated with the tooling forwards-compatible with a later version that removes container-interop completely.

Essentially, all this does is **get rid of the warnings about installing container-interop**. A new major version is still required for removing container-interop usage, as we will need to update typehints against the container-interop `ContainerInterface`.


weierophinney:
Ping @Ocramius

weierophinney:
One thing brought up around solutions like this is that the new preloading features in PHP 7.4 may not work correctly; I'll do some tests and provide the results later.

Ocramius:
Discussed briefly in chat:

 * will break 7.4 preloading
 * we need tests for preloading (no idea how that would be designed, sorry :S )
 * possible solution for now could be to import `ContainerInterface`, `@deprecate` it ourselves and `{"provides": {"container-interop/container-interop": "1.99.99"}}` to remove dependency

src/AbstractPluginManager.php
```diff
@@ -49,21 +48,11 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
...
     {
-        if ($configInstanceOrParentLocator instanceof PsrContainerInterface
-            && ! $configInstanceOrParentLocator instanceof ContainerInterface
-        ) {
-            /**
-             * {@see \Laminas\ServiceManager\Factory\FactoryInterface} typehints
-             * against interop container and as such cannot accept non-interop
-             * psr container. Decorate it as interop.
-             */
-            $configInstanceOrParentLocator = new PsrContainerDecorator($configInstanceOrParentLocator);
```
michalbundyra: We can deprecate now `PsrContainerDecorator`.

src/AbstractPluginManager.php
```diff
@@ -49,21 +48,11 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
      * factories; for $config, {@see \Laminas\ServiceManager\ServiceManager::configure()}
      * for details on its accepted structure.
      *
-     * @param null|ConfigInterface|ContainerInterface|PsrContainerInterface $configInstanceOrParentLocator
+     * @param null|ConfigInterface $configInstanceOrParentLocator
```
michalbundyra: We should keep here `ContainerInterface` (`Interop`) as it is `Psr` through the alias.

src/AbstractPluginManager.php
```diff
@@ -8,9 +8,8 @@
 
 namespace Laminas\ServiceManager;
 
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
```
michalbundyra: Per the previous comment, I think we should keep here `Interop` not `Psr`. We are using `Psr` through the alias, so it should be fine.

boesing:
Is this meant to be a forward-compatible or is this targeting the SM 4.0?
If the latter, why would we `class_alias` interop at all?

weierophinney:
@boesing This aims to be forward compatible. The main idea is to no longer require container-interop, and allow users to use PSR-11 typehints instead of container-interop typehints where possible.

I plan to incorporate feedback on this soon, and will ping the TSC when it's ready for another review.

.travis.yml
```diff
@@ -39,6 +39,18 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
+    - php: 7.3
+      env:
+        - DEPS=lowest
```
samsonasik, line 44: lowest on php 7.3 can be removed to match against develop

weierophinney: develop should do both lowest and latest... if it's not, we'll need to add it.

samsonasik: I commented at https://github.com/weierophinney/laminas-servicemanager/pull/1#issuecomment-626799552 about the reason lowest at php 7.3 removed

weierophinney:
Okay, now that Travis is finally green...

@laminas/technical-steering-committee — This is ready to review again. I've changed the scope, per concerns about opcache preloading, to do the following:

- Instead of using `class_alias`, I've imported the container-interop interfaces so we can provide them _within the package_.

- I've modified the `AbstractPluginManager` to no longer decorate PSR-11 containers. There is no need, as container-interop and PSR-11 are interchangeable here, and we now guarantee that container-interop fulfills the PSR-11 contract (since we provide the versions that extend PSR-11). As such, I marked the `PsrContainerDecorator` as deprecated.

- I've modified the `FactoryCreator` to generate a **generic factory** — i.e., one that does not explicitly implement `FactoryInterface`. This allows factories it generates to typehint against PSR-11, making them forwards-compatible.

This does, mean, however, that implementations of our interfaces must still typehint against container-interop, as changing their typehints would break backwards compatibility. My plan is that for 4.0, we can also provide tooling to update any factories implementing our interfaces.

.travis.yml
weierophinney: I was able to workaround it by adding some logic to use later versions of phpunit and ocramius/proxy-manager. Since these are _testing_requirements, and not _library_ requirements, we can use later versions and still be testing the functionality properly. As you'll see now, tests are green!

src/AbstractPluginManager.php
weierophinney: By having container-interop inlined in the package, we now ensure that anything container-interop is also PSR-11, and since we only consume the PSR-11 interfaces within the plugin manager, I think typehinting against it here is acceptable, and signals the intent for future versions.

src/AbstractPluginManager.php
weierophinney: Done

Xerkus:
Sorry, but I fail to see how is this different from just requiring interop package itself?

Point of `PsrContainerDecorator` is not for the plugin manager but for the user factories that typehint against Interop container and can't accept Psr container that is not Interop. It cannot be removed without BC break as without alias they are not the same.

weierophinney:
@Xerkus —

> Sorry, but I fail to see how is this different from just requiring interop package itself?

In large part, it's to reduce the number of people:

a) asking how to get rid of the warning due to installing container-interop, and
b) wondering why we haven't adopted PSR-11 fully yet, and thinking we're "dying" as a project because of this one, trivial thing. (Fact is, we _have_ fully adopted PSR-11. We just haven't fully divested ourselves of container-interop, due to BC issues.)

I know I've had to answer these questions several dozen times or more, and this is a stop-gap so we can focus on removing it entirely. My original plan to use aliasing worked well... until we started looking at opcache preloading, and it broke things. And, frankly, even with that plan, we'd still need to do the same migration tooling for v4, as folks would still have the container-interop signatures in place.

Regarding:

> Point of `PsrContainerDecorator` is not for the plugin manager but for the user factories that typehint against Interop container and can't accept Psr container that is not Interop. It cannot be removed without BC break as without alias they are not the same.

That's exactly the sort of feedback I was looking for, as I'd forgotten that aspect. I'll revert those changes.

@geerteltink —

> But I rather like to see a 4.0 version with all container-interop stuff removed completely.

That'll be the next step after these changes are released (pending the rollback based on the feedback of @Xerkus)

composer.json
```diff
@@ -78,6 +78,7 @@
         "upload-coverage": "coveralls -v"
     },
     "replace": {
+        "container-interop/container-interop": ">=1.2.0",
```
michalbundyra, line 81: Isn't this gonna be a problem if I have installed container-interop < 1.2.0 ?

weierophinney: We already conflicted with versions less than that. :wink:

weierophinney:
I'm going to close this without merging.

My original goal was to create a patch that made existing code forwards compatible by aliasing the container-interop interfaces to the PSR-11 interfaces. However, this will not work for a number of reasons:

- It breaks when used in conjunction with opcache preloading.
- Users who were typehinting against the container-interop interfaces (e.g., because they implemented our own factory interfaces) would end up with code breakage at the next major release anyways, unless we also provided a tool to update their existing code to typehint against PSR-11 interfaces — which we cannot do due to the previous point.

Since the patch cannot meet the goals I set out to achieve, I think it makes more sense to work towards a new major version that both removes the dependency on container-interop, as well as provides tooling to fix code that uses those interfaces (either via implementation or typehinting).

The API we provide does not necessarily need to change, which would mean we could update code that depends on laminas-servicemanager to allow either v3 or v4 releases; if they provide any interface implementations (e.g., `FactoryInterface`, `DelegatorFactoryInterface`, etc.), we can remove the implementation aspect, as all that is really required in the majority of cases is an `__invoke()` method (as demonstrated with Mezzio).

Xerkus:
@weierophinney factory interfaces are nice to have for user convenience, especially when extra functionality is provided eg abstract factories with `canCreate()`.

Going forward, I think:
- factory interfaces provided here should be deprecated and removed with interop dependency in the next major
- independent package introduced that provides only PSR-11 container targeting factory interfaces.
  Common factories also deprecated and moved there, `InvokableFactory` is the prime candidate that is used with other containers in Mezzio.
- As you mentioned, for backward and forward compatibility service manager consumers would drop interface usage. Optionally, they can adopt independent interfaces if those are introduced.

Xerkus:
Possible issue to keep in mind:
Dropping interfaces and changing to psr container typehint in factories provided by components across the framework might cause some breakage if those factories were extended and overridden in user code as we did not mark factories final.

Such change would need to be communicated with extra emphasis in every affected component.

weierophinney:
@Xerkus I'm working on a plan now, and surveying all packages we ship that have any laminas-servicemanager interface implementations so that I can determine scope and impact. It's slow-going (there are a LOT of factory implementations, as it turns out), but I'm already formulating some ideas for how we can create forwards-compatible releases to make transitioning easier.

PR #32, opened 2020-02-29
Update doc configuring-the-service-manager

wotent:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

ServiceManager's configure method will convert all invokables into factories,
so we should use InvokableFactory for better performance.


weierophinney:
Thanks, @wotent !

PR #35, opened 2020-04-18
update exception messages

pine3ree:
Just txt changes to have clearer messages. Not sure that in ConfigAbstractFactory `Service messages` was a typo and intended to be "Service dependencies".

There are 2 unrelated cs warnings: https://travis-ci.org/github/pine3ree/laminas-servicemanager/builds/676703964, the first can be fixed, the second causes problems with the FactoryCreatorCommand.


weierophinney:
Thanks, @pine3ree !

PR #40, opened 2020-05-11
Alphabetize import statements in generated factories

weierophinney:
Previously, the imports were _mostly_ alphabetized. However, with the switch to Laminas, the container import was now happening incorrectly with regards to the factory interface, causing test failures when doing comparisons between expected results and actual results. Additionally, it makes sense to alphabetize to ensure we follow our own CS.

This patch updates the factory generation code to alphabetize import statements before injecting them in the template.

The work is done on the develop branch as that was the branch exhibiting the test failure behaviors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,6 @@ matrix:
         - DEPS=latest
     - php: 7.3
       env:
-        - DEPS=lowest
-    - php: 7.3
-      env:
         - DEPS=latest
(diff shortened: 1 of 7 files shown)


samsonasik:
@weierophinney I created PR to your PR branch https://github.com/weierophinney/laminas-servicemanager/pull/1 to fix the phpunit error

PR #43, opened 2020-06-05
#3 Ported 279 from ZF repository - fixes merging of duplicate delegator factory definitions

### LINKED ISSUE #3: Hotfix #279
Fix issue #279 

---

Originally posted by @thomasvargiu at https://github.com/zendframework/zend-servicemanager/pull/280

GeeH:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
Bug fix from old repostiory at https://github.com/zendframework/zend-servicemanager/pull/280

Thanks @thomasvargiu 

Fixes #3

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -748,13 +748,13 @@ class ServiceManager implements ServiceLocatorInterface
     private function mergeDelegators(array $config)
     {
         foreach ($config as $key => $delegators) {
-            if (!isset($this->delegators[$key])) {
+            if (! isset($this->delegators[$key])) {
                 $this->delegators[$key] = $delegators;
                 continue;
             }
 
             foreach ($delegators as $delegator) {
-                if (!in_array($delegator, $this->delegators[$key], true)) {
+                if (! in_array($delegator, $this->delegators[$key], true)) {
                     $this->delegators[$key][] = $delegator;
                 }
             }
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -21,7 +21,7 @@ use ProxyManager\FileLocator\FileLocator;
 use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
 use ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy;
 use Psr\Container\ContainerInterface;
-use Zend\Stdlib\ArrayUtils;
+use Laminas\Stdlib\ArrayUtils;
 
 use function array_intersect;
 use function class_exists;
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -748,7 +748,7 @@ class ServiceManager implements ServiceLocatorInterface
     private function mergeDelegators(array $config)
     {
         foreach ($config as $key => $delegators) {
-            if (! isset($this->delegators[$key])) {
+            if (! array_key_exists($key, $this->delegators)) {
                 $this->delegators[$key] = $delegators;
                 continue;
             }
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -21,6 +21,7 @@ use ProxyManager\FileLocator\FileLocator;
 use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
 use ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy;
 use Psr\Container\ContainerInterface;
+use Zend\Stdlib\ArrayUtils;
 
 use function array_intersect;
 use function array_merge_recursive;
@@ -363,7 +364,7 @@ class ServiceManager implements ServiceLocatorInterface
         }
 
         if (isset($config['delegators'])) {
-            $this->delegators = array_merge_recursive($this->delegators, $config['delegators']);
+            $this->mergeDelegators($config['delegators']);
         }
 
         if (isset($config['shared'])) {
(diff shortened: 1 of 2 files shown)


GeeH:
Closes #3

src/ServiceManager.php
```diff
@@ -737,6 +737,32 @@ private function createLazyServiceDelegatorFactory()
...
+     * It works with strings and class instances.
+     * It's not possible to de-duple anonymous functions
+     *
+     * @param string[][]|Factory\DelegatorFactoryInterface[][] $config
+     * @return string[][]|Factory\DelegatorFactoryInterface[][]
+     */
+    private function mergeDelegators(array $config)
+    {
+        foreach ($config as $key => $delegators) {
+            if (! isset($this->delegators[$key])) {
```
Ocramius, line 751: This should probably be an `array_key_exists()` check to prevent `isset()` semantics (`isset()` can work with undefined variables, which may happen when refactoring at a later point in time)

src/ServiceManager.php
```diff
@@ -737,6 +737,32 @@ private function createLazyServiceDelegatorFactory()
...
+                $this->delegators[$key] = $delegators;
+                continue;
+            }
+
+            foreach ($delegators as $delegator) {
+                if (! in_array($delegator, $this->delegators[$key], true)) {
```
Ocramius, line 757: This `in_array()` is repeated multiple times: potentially an issue as the array grows.

I don't think it's an issue, since collisions are rare, but worth noting.

Up to you if it needs to be adjusted or not.

test/ServiceManagerTest.php
```diff
@@ -364,4 +365,51 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
 
         self::assertFalse($serviceManager->has('Alias'));
     }
+
+    /**
+     * Hotfix #3
```
Ocramius, line 370: Maybe `@group #3` would be better - not really a hotfix anymore :)

test/ServiceManagerTest.php
```diff
@@ -364,4 +365,51 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
...
+     * @see https://github.com/laminas/laminas-servicemanager/issues/3
+     */
+    public function testConfigureMultipleTimesAvoidsDuplicates()
+    {
+        $delegatorFactory = function (
+            ContainerInterface $container,
+            $name,
```
Ocramius, line 377: Should probably have `string` here

test/ServiceManagerTest.php
```diff
@@ -364,4 +365,51 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
 
         self::assertFalse($serviceManager->has('Alias'));
     }
+
+    /**
+     * Hotfix #3
+     * @see https://github.com/laminas/laminas-servicemanager/issues/3
+     */
+    public function testConfigureMultipleTimesAvoidsDuplicates()
```
Ocramius, line 373: Probably better naming needed: `testConfiguringADelegatorMultipleTimesDoesNotLeadToDuplicateDelegatorCalls()`

test/ServiceManagerTest.php
```diff
@@ -364,4 +365,51 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
...
+                    return new InvokableObject();
+                },
+            ],
+            'delegators' => [
+                'Foo' => [
+                    $delegatorFactory,
+                    LazyServiceFactory::class,
+                ],
+            ],
+            'lazy_services' => [
```
Ocramius, line 399: This test is verifying both the `lazy_services` and the `delegator` changes: should probably be split up into two tests

test/ServiceManagerTest.php
```diff
@@ -364,4 +365,51 @@ public function testResolvedAliasNoMatchingAbstractFactoryReturnsFalse()
...
+            ],
+        ];
+
+        $serviceManager = new ServiceManager($config);
+        $serviceManager->configure($config);
+
+        /** @var InvokableObject $instance */
+        $instance = $serviceManager->get('Foo');
+
+        self::assertInstanceOf(InvokableObject::class, $instance);
```
Ocramius, line 412: This assertion is not needed (based on `/** @var */` right above)

test/ServiceManagerTest.php
GeeH: I've copied this code from the original repo and cast my eye over it for sanity but I absoltely could add type declarations if we think that's appropriate. My thought was to keep it in line with the existing code which doesn't have type declarations but it's a good chance to start moving in that direction if we think it's a good time to start.

How would we feel about some methods having types and others not?

test/ServiceManagerTest.php
GeeH: The `@var` is only an IDE hint and not actually checking that the instance returned is the correct class tho, right?

test/ServiceManagerTest.php
GeeH: I think this is OK; it's in the scope of the bugfix imo.

test/ServiceManagerTest.php
weierophinney: If we're bumping the PHP version to v3, I'd argue _add them everywhere_ in the test suite.

Ocramius: Yeh, but then not in this patch.

We can run psalter manually in a different patch.

Discussed this with @GeeH and it should be fine as-is for now

test/ServiceManagerTest.php
Ocramius: It will likely be picked up by static analysis later, but we can ignore it for now :+1:

GeeH:
Ready for re-review/merge thanks both.

src/ServiceManager.php
```diff
@@ -21,9 +21,9 @@
 use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
 use ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy;
 use Psr\Container\ContainerInterface;
+use Zend\Stdlib\ArrayUtils;
```
samsonasik, line 24: /s/Zend/Laminas

GeeH:
ping @Ocramius ?

Issue #44, opened 2020-06-12
Add PHP 7.3 type declarations


GeeH:
- add `declare(strict_types=1)` to all src and test files
 - add parameter type declarations where appropriate
 - add return type declarations where appropriate
 - rejoice in a job well done

ppaulis:
@GeeH need some help on this?
I took a quick look at the dependencies. Not everything seems to be PHP8-ready yet, but I could help with the PSR-12 and the property types in the meantime.

nusphere:
i also could help to improve the PSR-12 Coding Style and Type-Hint.

nusphere:
@GeeH @ppaulis 
is there currently someone who is actively working on it?

is it possible that I can divide myself into it? Because doing something twice is more or less pointless.
But since the issue has not been further developed since June 18th ... what is the current status?
or does it make sense to simply fork the current 4.0.0 branch and off you go?

ppaulis:
@nusphere I'm waiting for some maintainer input here to avoid doing the work at the same time than someone else.

nusphere:
@ppaulis 

Thank you for your feedback. in the meantime i have tried other components, as there is basically a lot to do.

Doing double the job makes absolutely no sense.

I am happy to help with this task, but I am now trying to create the basis for such a refactoring by helping to provide the GitHub Actions in the components.

GeeH:
Hi, huge apologies, real life took over. I'm happy to get back involved here but would you be able to roundup what you need from me please @nusphere ?

boesing:
@GeeH In the meantime, we dropped v4.0.0 as having a new major version in the service manager would lead to a huge amount of work in almost every other component which works with the servicemanager directly (e.g. due to the usage of `AbstractPluginManager`).

Another thing is a proper migration path for the `psr/container` BC break.

Thus said, I don't think it makes sense to approach type-safety or PHP 7.4 type declarations without having a proper strategy on how we want to proceed with this component.

fabiang:
Just a side note for the one who will implement this, the return-type of `Laminas\ServiceManager\Factory\FactoryInterface::__invoke()` is currently declared as `@return object`, but even in real-world application `__invoke()` could return everything (even scalars). 
So `mixed` is the correct return-type here. One example is `Laminas\Mvc\Service\ConfigFactory::__invoke()` which returns `iterable`.

boesing:
Closed with #191

PR #46, opened 2020-06-12
#44 add strict type declarations

GeeH:
### Description

Adds strict type declarations to all files in `src/`, `bin/` and `tests/` as per #44 

Fixex #44

### DIFF
--- a/test/Tool/FactoryCreatorCommandTest.php
+++ b/test/Tool/FactoryCreatorCommandTest.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * @see       https://github.com/laminas/laminas-servicemanager for the canonical source repository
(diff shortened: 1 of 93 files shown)


test/Tool/FactoryCreatorCommandTest.php
```diff
@@ -16,11 +18,18 @@
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
 
+use Prophecy\Prophecy\ObjectProphecy;
 use function file_get_contents;
 use function sprintf;
 
 class FactoryCreatorCommandTest extends TestCase
 {
+    /** @var ObjectProphecy|ConsoleHelper */
```
Ocramius, line 27: This should be an intersection type

Ocramius: This can/should be fixed by using psalter later on

PR #49, opened 2020-06-15
Removed deprecated interfaces

GeeH:
Deleted the 4 deprecated interfaces in preperation for the 4.0.0 release.

`AbstractFactoryInterface`
`DelegatorFactoryInterface`
`FactoryInterface`
`InitializerInterface`


Ocramius (approved):
I think this needs to be first passed to review in the @laminas/technical-steering-committee meeting - don't see problems with it otherwise

PowerKiKi:
According to https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2021-02-01-TSC-Minutes.md#future-of-laminas-servicemanager, this PR should be rejected and a new one should be created from scratch.

PR #50, opened 2020-06-18
Updated mimumum PHP version to 7.4 and tried to add 8.0 to CI

GeeH:
Signed-off-by: Gary Hockin <gary@hock.in>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | yes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,19 +12,19 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.3
+    - php: 7.4
       env:
         - DEPS=lowest
-    - php: 7.3
+    - php: 7.4
       env:
         - DEPS=latest
         - CS_CHECK=true
         - BENCHMARKS=true
         - TEST_COVERAGE=true
-    - php: 7.4
+    - php: 8.0
       env:
         - DEPS=lowest
-    - php: 7.4
+    - php: 8.0
       env:
         - DEPS=latest
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,10 @@ matrix:
         - CS_CHECK=true
         - BENCHMARKS=true
         - TEST_COVERAGE=true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -12,19 +12,19 @@ env:
...
         - DEPS=lowest
-    - php: 7.3
+    - php: 7.4
       env:
         - DEPS=latest
         - CS_CHECK=true
         - BENCHMARKS=true
         - TEST_COVERAGE=true
-    - php: 7.4
+    - php: 8.0
```
Ocramius, line 24: We should have this in allowed failures

.travis.yml
```diff
@@ -12,19 +12,19 @@ env:
...
       env:
         - DEPS=lowest
-    - php: 7.4
+    - php: 8.0
```
Ocramius, line 27: We should have this in allowed failures

Ocramius (changes requested):

.travis.yml
michalbundyra: this is not working yet on travis, we have to use nightly if we want test on PHP 8.0

and of course we need much more changes in the travis to allow it to happen....

.travis.yml
GeeH: We're doing it on stream at https://twitch.tv/spabby - come and tell us how.

Ocramius (approved):
:ship:

michalbundyra:
@GeeH @Ocramius 
what's the point of adding nightly build if these are not even configured properly?
See travis build:

> - This package requires php ^7.4 but your PHP version (8.0.0-dev) does not satisfy that requirement.

We need to change the process there to install dependencies when using nightly... Otherwise it has no sense at all.

PR #47, opened 2020-06-13
[WIP] Bump PHPUnit:^9

sasezaki:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Currently, I feel needs to remove deprecations (eg.  in AbstractPluginManager) before update to PHPUnit 9.

Because, `assertAttributeSame()` is used to check raising E_USER_DEPRECATED.

Introduced BackportAssertionsTrait should be removed before merge.

PR #56, opened 2020-10-10
add phpunit9 support

gennadigennadigennadi:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description
I jus took the work from https://github.com/laminas/laminas-servicemanager/pull/47 and finished the missing parts


gennadigennadigennadi:
@sasezaki thank you for your work! 

There is still the problem with the phpcsfixer throwing warning because of the {} for accessing arrays.

test/AbstractPluginManagerTest.php
```diff
@@ -34,7 +34,9 @@
  */
 class AbstractPluginManagerTest extends TestCase
 {
+    use \Prophecy\PhpUnit\ProphecyTrait;
```
samsonasik: please move to use statement

gennadigennadigennadi: done.

test/BackportAssertionsTrait.php
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Backport Assertion Already removed latest PHPUnit.
+ *
+ * @see https://github.com/sebastianbergmann/phpunit/issues/3339
+ *
+ * borrowed from
+ * https://github.com/sebastianbergmann/phpunit/blob/7.5.9/src/Framework/Assert.php#L1408
+ */
+trait BackportAssertionsTrait
```
froschdesign: The actual problem is only moved but not fixed.
Please update the related tests instead of introducing new code which must be maintained additionally.

There is a moderate number of tests:

* https://github.com/laminas/laminas-servicemanager/search?q=assertAttributeSame
* https://github.com/laminas/laminas-servicemanager/search?q=assertAttributeEquals
* https://github.com/laminas/laminas-servicemanager/search?q=assertInternalType

Thanks in advance!

gennadigennadigennadi: I have replaced them. 

- assertInternalType => assertIsArray.
- for AssertEquals  I have added getCreationContext() inside the TestAssets.
- for assertAttributeSame I have added anonymous classes to get access to the private properties.

gennadigennadigennadi:
@froschdesign and @samsonasik is their something else I should still do?

samsonasik (approved):

boesing:
Thanks, @gennadigennadigennadi!

PR #64, opened 2020-12-02
Backport 4.x

boesing:


.travis.yml
```diff
@@ -6,45 +6,43 @@ cache:
...
         - DEPS=lowest
-    - php: 7.1
+    - php: 7.4
       env:
         - DEPS=latest
-        - BENCHMARKS=true
-        - CS_CHECK=true
-        - TEST_COVERAGE=true
-    - php: 7.2
+    - php: nightly
```
snapshotpl: Travis support `8.0` now

.travis.yml
```diff
@@ -6,45 +6,43 @@ cache:
...
       env:
         - DEPS=lowest
-    - php: 7.2
+        - COMPOSER_ARGS="$COMPOSER_ARGS --ignore-platform-reqs"
```
snapshotpl: Are we ok with `ignore-platform-reqs` ? I think it's risky to use something that is not right tested. Which dependencies require this flag?

.travis.yml
```diff
@@ -6,45 +6,43 @@ cache:
...
+    - php: nightly
       env:
         - DEPS=latest
+        - COMPOSER_ARGS="$COMPOSER_ARGS --ignore-platform-reqs"
 
 before_install:
   - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
+  # Restore composer back to v1
+  - composer self-update --1
```
snapshotpl: Why?

boesing: Due to incompatibility with `ocramius/package-versions` and I dont want to spend too much time in this to fix as it works with composer v1 without any issue.

https://travis-ci.com/github/laminas/laminas-servicemanager/jobs/452925196#L456

.travis.yml
boesing: Since 3rd-party dependencies do not provide proper constraints, this is still needed.
It is risky but thats how it is for now and thats how it is in almost every other laminas/mezzio package due to the lack of support in other libs.

.travis.yml
boesing: Since when? Tested that 3 days ago and it did not.
Any official resource to verify your statement before I start changing `.travis.yml` twice?

boesing: https://travis-ci.community/t/php-8-0-missing/10132/23

composer.json
```diff
@@ -23,33 +23,35 @@
...
         "psr/container-implementation": "^1.0"
     },
+    "conflict": {
+        "zendframework/zend-code": "<3.3.1",
+        "laminas/laminas-code": "<3.3.1"
+    },
     "suggest": {
-        "laminas/laminas-stdlib": "laminas-stdlib ^2.5 if you wish to use the MergeReplaceKey or MergeRemoveKey features in Config instances",
-        "ocramius/proxy-manager": "ProxyManager 1.* to handle lazy initialization of services"
+        "laminas/laminas-stdlib": "laminas-stdlib ^2.7.7 | ^3.1 if you wish to use the MergeReplaceKey or MergeRemoveKey features in Config instances",
```
froschdesign: The "require" section already contains laminas-stdlib therefore this line can be removed.

src/AbstractFactory/ConfigAbstractFactory.php
```diff
@@ -64,8 +72,8 @@ public function __invoke(\Interop\Container\ContainerInterface $container, $requ
 
         $serviceDependencies = $dependencies[$requestedName];
 
-        if ($serviceDependencies !== array_values(array_map('strval', $serviceDependencies))) {
-            $problem = json_encode(array_map('gettype', $serviceDependencies));
+        if ($serviceDependencies !== array_values(array_map('\strval', $serviceDependencies))) {
```
weierophinney: Functions and classes referenced inside of strings are always considered globally qualified; there's no reason to prefix them.

src/Exception/ContainerModificationsNotAllowedException.php
```diff
@@ -10,9 +12,24 @@
...
 /**
  * @inheritDoc
  */
 class ContainerModificationsNotAllowedException extends DomainException implements ExceptionInterface
 {
+    /**
+     * @param string $service Name of service that already exists.
+     * @return self
+     */
+    public static function fromExistingService($service)
```
weierophinney: Since this is a new method, and we're updating to PHP 7.4 as the minimum supported version, provide explicit argument and return typehints here, please.

src/Exception/CyclicAliasException.php
```diff
@@ -8,11 +10,41 @@
...
+use function sprintf;
+
 class CyclicAliasException extends InvalidArgumentException
 {
+    /**
+     * @param string   $alias conflicting alias key
+     * @param string[] $aliases map of referenced services, indexed by alias name (string)
+     * @return self
+     */
+    public static function fromCyclicAlias($alias, array $aliases)
```
weierophinney: Provide all argument and return type hints on new methods, please.

src/Exception/InvalidArgumentException.php
```diff
@@ -9,10 +11,44 @@
...
+            InitializerInterface::class,
+            is_object($initializer) ? get_class($initializer) : gettype($initializer)
+        ));
+    }
+
+    /**
+     * @param mixed $abstractFactory
+     * @return self
+     */
+    public static function fromInvalidAbstractFactory($abstractFactory)
```
weierophinney: Typehints, please.

src/Tool/FactoryCreator.php
```diff
@@ -39,6 +56,11 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
 
 EOT;
```
weierophinney: Since we're bumping the minimum supported PHP version... PHP 7.3+ allow you to indent heredocs and nowdocs. The closing marker (in this case, `EOT;`) defines the amount of indentation that will be stripped from the start of each line of the heredoc/nowdoc. I've done this in a few other components, and it really helps improve readability.

src/Tool/FactoryCreator.php
```diff
@@ -110,7 +135,8 @@ function (ReflectionParameter $argument) {
         }
 
         return array_map(function (ReflectionParameter $parameter) {
```
weierophinney: This should have a return typehint of `?string`.

src/Tool/FactoryCreator.php
```diff
@@ -140,4 +166,13 @@ private function createArgumentString($className)
                 );
         }
     }
+
+    private function createImportStatements(string $className): string
+    {
+        $imports = array_merge(self::IMPORT_ALWAYS, [$className]);
+        sort($imports);
+        return implode("\n", array_map(function ($import) {
```
weierophinney: This should have a return typehint of `string`.

test/AbstractFactory/ConfigAbstractFactoryTest.php
```diff
@@ -185,8 +187,8 @@ public function testExceptsWhenConfigNotSet()
     {
         $abstractFactory = new ConfigAbstractFactory();
         $serviceManager = new ServiceManager();
-        self::expectException(ServiceNotCreatedException::class);
-        self::expectExceptionMessage('Cannot find a config array in the container');
+        $this->expectException(ServiceNotCreatedException::class);
+        $this->expectExceptionMessage('Cannot find a config array in the container');
```
weierophinney, line 191: Arggghhh!

We need to likely decide on which syntax we use, so that we don't have commits like this where we flip between styles.

Prior to the last year or so, we were consistently using instance access; in the past year, particularly with some of the updates to add Psalm support, there have been a number of switches to static access. We need to pick one and be consistent everywhere.

For this patch, I'd suggest leaving these alone, as they add a lot of extra noise to the patch.

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -16,13 +18,103 @@
...
+
+    /**
+     * Test data provider for testFromCyclicAlias
+     *
+     * @return string[][]|string[][][]
+     */
+    public function cyclicAliasProvider()
+    {
+        return [
+            [
```
weierophinney: Please give these names (e.g. "a -> a", "a -> b -> a", etc.).

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -16,13 +18,103 @@
...
+        self::assertInstanceOf(CyclicAliasException::class, $exception);
+        self::assertSame($expectedMessage, $exception->getMessage());
+    }
+
+    /**
+     * Test data provider for testFromCyclicAlias
+     *
+     * @return string[][]|string[][][]
+     */
+    public function cyclicAliasProvider()
```
weierophinney: Return typehint, please.

test/Exception/CyclicAliasExceptionTest.php
```diff
@@ -16,13 +18,103 @@
...
+
     /**
-     * @dataProvider aliasesProvider
+     * @dataProvider cyclicAliasProvider
      *
+     * @param string   $alias, conflicting alias key
      * @param string[] $aliases
      * @param string   $expectedMessage
+     */
+    public function testFromCyclicAlias($alias, array $aliases, $expectedMessage)
```
weierophinney: Typehints, please.

weierophinney (changes requested):
Main comments:

- We're bumping the minimum supported PHP version to 7.4. So all NEW methods should have full typehinting in place; this includes closures.
- Add indenting to heredoc/nowdoc definitions (I have a comment on that later).
- Don't switch between `$this->` and `self::` in the tests; it's a lot of noise, and we really need to have a discussion about this in the TSC to decide which one we're going to use going forward. For now, instance usage is consistent with other components, and reduces the size of this patch considerably.

weierophinney (approved):

PR #67, opened 2020-12-30
Inspection

zakirullin:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

Static time inspection is a quite nice thing to have (especially if you don't favor the `AoT`). 
At the moment it could detect following defects:

- Circular dependency
- Cyclic alias
- Missing factory
- Factory autoload failure
- Unresolvable params/scalars
- Scalar type/name mismatch
- Many other scenarios of misconfiguration 

image

Can also be paired with `laminas-di` & `AutowireFactory` (an additional resolver is needed, goes off the scope of current PR, a delegator will be provided in `laminas-di` scope). 

P.S. Tests & `Psalm` & `laminas-cli` &  other small fixes will follow.

P.P.S. Maybe this is a good candidate for a separate component, like `laminas-servicemanager-inspector`, given that the changes are quite isolated.


zakirullin:
So far it analyses both `ReflectionBasedAbstractFactory` & `AutowireFactory`.
As for custom factories, I did an AST-based dependencies resolver, it works quite good, but of course has some false positive/negative issues, which is not applicable. With all that complication and outcome in mind - it's not worth it.

weierophinney:
@zakirullin We need you do provide sign-off on your commits; you can [read how to do so here](https://github.com/laminas/.github/blob/master/CONTRIBUTING.md#adding-signoffs-after-the-fact).

zakirullin:
Hm, AST-based factory analyzer yielded fairly impressive results.
Just for fun did a `rector-php` node visitor, which is looking for `ContainerInterface::get($serviceName)` calls in userland factories.
It works great in 98% cases. But that's not a subject to include in the first place.

boesing:
I think we can close this in favor of https://github.com/laminas/laminas-servicemanager-inspector

PR #71, opened 2021-01-10
Regression fix regarding initializers

### LINKED ISSUE #70: All but the first configured initializers are ignored
### BC Break Report (or bug?)

Broken version: 3.6.0
Working version: 3.5.1
Commit: https://github.com/laminas/laminas-servicemanager/commit/d0b38fb4629d8bf7111cb0533f6a512b1fd10923#diff-c88e8922673ece50d04aef21f4dc95b87bd465868eb073480467f87e626e3b0aR541 merged by 0a6c0ffc88a91720b3199b91d0374eb0f20dcfd0

#### Summary

This code in the ServiceManager->resolveInitializers() method:
```php
            if (is_callable($initializer)) {
                $this->initializers[] = $initializer;
                return;
            }
```
causes to stop processing initializers after the first one. There should be `continue`, not `return`.

This bug probably caused by refactoring: this lines were in separate function `resolveInitializer()`, where `return` was ok. But then code had copy-pasted into foreach loop in base function, without altering `return` to `continue`.

#### Previous behavior

All passed initializers are processed and used.

#### Current behavior

Only the first initializer are processed and used.

#### How to reproduce

Just add two initializers to configuration:

```php
class Test {}

$sm = new \Laminas\ServiceManager\ServiceManager([
    'invokables' => [
        Test::class => Test::class
    ],
    'initializers' => [
        function() { echo "First\n"; },
        function() { echo "Second\n"; },
    ],
]);
$sm->get(Test::class);
```
Expected output:
```
First
Second
```
Actual output:
```
First
```

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

While backporting changes from 4.x, a change slipped through our fingers and thus introduced an annoying BC break.

Fixes #70

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -554,7 +554,7 @@ class ServiceManager implements ServiceLocatorInterface
 
             if (is_callable($initializer)) {
                 $this->initializers[] = $initializer;
-                return;
+                continue;
             }
 
             throw InvalidArgumentException::fromInvalidInitializer($initializer);
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -446,4 +446,34 @@ class ServiceManagerTest extends TestCase
         $this->assertInstanceOf(InvokableObject::class, $instance);
         $this->assertSame(1, $instance->getOptions()['inc']);
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


Ocramius (approved):
LGTM 👍

Issue #68, opened 2021-01-07
Overwriting an invokable with a factory does not work


alexz707:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 3.5.1

#### Summary

The config get's merged not as expected.
I have a vendor's config in Module.php like this:
```PHP
invokables' => [
    'lmcuser_register_form_hydrator' => \Laminas\Hydrator\ClassMethodsHydrator::class,
]
```
I want to overide this one with a factory in my module.config.php:

```PHP
    'service_manager' => [
        'factories' => [
            'lmcuser_register_form_hydrator' => DoctrineHydratorfactory::class,
        ],
```
#### Current behavior

If I try to get the 'lmcuser_register_form_hydrator' from ServiceManager I get the ClassMethodsHydrator::class.

#### How to reproduce

Just overwrite a invokable (which gets aliased) with a factory and try to get it with dem sm.

#### Expected behavior
I would expect to get an Object from the DoctrineHydratorfactory::class which would give me in my case a DoctrineObject.

I think it has to do with the way invokables get aliased and then merged with the other config.
Maybe the array_merge arguments just have to be switched?

Servicemanager.php on line 325
```PHP
            if (! empty($aliases)) {
                $config['aliases'] = (isset($config['aliases']))
                    ? array_merge($config['aliases'], $aliases)
                    : $aliases;
            }
```

syberon:
Today i found the same bug in version 3.6.1 - all overwritten invokables are ignored.

The reason is in ServiceManager.php on line 376:
```
if (! empty($config['aliases'])) {
    $this->aliases = $config['aliases'] + $this->aliases;
    $this->mapAliasesToTargets();
} elseif (! $this->configured && ! empty($this->aliases)) {
    $this->mapAliasesToTargets();
}
```

This assignment overwrites all custom declared invokables by default values. To fix this bug this line should be:
`$this->aliases = $this->aliases + $config['aliases'];`

Ocramius:
@syberon can you produce a failing test case, please?
Should be easy to fix once that is available.

driehle:
I am having the same issue, related with the `neilime/twbs-helper-module` library. The issue appears with service manager 3.6.0 and 3.6.1, but not with 3.5.1. So for now I downgraded with `composer require laminas/laminas-servicemanager:"~3.5.0"`.

driehle:
Not a 100% sure, but I might have found a suitable test case for @syberon's problem:

[...]

PR #74, opened 2021-01-17
Fix invokables merging precedence in `3.6.x`

driehle:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| New Feature   | no

### Description

This PR fixes a BC break introduced in Laminas service manager 3.6.x. When invokables are defined with an alias, that alias used to override other aliases in 3.5.x series, which does not happen anymore in 3.6.x.  See PR #75, where I have backported the same test case to the 3.5.x branch.

See ticket #68 for a discussion on the bug. Affected projects are, for instance, neilime/twbs-helper-module#128 and LM-Commons/LmcUser#13.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -357,7 +357,10 @@ class ServiceManager implements ServiceLocatorInterface
         }
 
         if (isset($config['invokables']) && ! empty($config['invokables'])) {
-            $this->createAliasesAndFactoriesForInvokables($config['invokables']);
+            $newAliases = $this->createAliasesAndFactoriesForInvokables($config['invokables']);
+            // override existing aliases with those created by invokables to ensure
+            // that they are still present after merging aliases later on
+            $config['aliases'] = $newAliases + ($config['aliases'] ?? []);
         }
 
         if (isset($config['factories'])) {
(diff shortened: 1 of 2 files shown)


src/ServiceManager.php
```diff
@@ -357,7 +357,10 @@ public function configure(array $config)
         }
 
         if (isset($config['invokables']) && ! empty($config['invokables'])) {
-            $this->createAliasesAndFactoriesForInvokables($config['invokables']);
+            $newAliases = $this->createAliasesAndFactoriesForInvokables($config['invokables']);
+            // override existing aliases with those created by invokables to ensure
+            // that they are still present after merging aliases later on
+            $config['aliases'] = $newAliases + ($config['aliases'] ?? []);
```
Ocramius, line 363: I think `+` should be avoided here, since it has weird merging semantics: https://3v4l.org/mkf2r

Ocramius (changes requested):

src/ServiceManager.php
driehle: @Ocramius I don't think so. The only "weird" thing about the `+` operator for arrays is that values from the left operand take precedence over values from the right operand, which is different from the order of parameters in `array_merge`. PHP [documentation](https://www.php.net/manual/en/language.operators.array.php) says this clearly:

>   The + operator returns the right-hand array appended to the left-hand array; for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored.

See https://3v4l.org/0gZ0l for an example. Using the `+` operator would also be more consistent with [lines](https://github.com/laminas/laminas-servicemanager/blob/56da7b2d6f136aa0b9f426f98fc4bd5a03db48b1/src/ServiceManager.php#L356) 356, 367, 375 and 379.

Ocramius: Hmm, fairly sure I saw some weird behavior some weeks ago. I'm ok with this scope after this convo, so I'd say :ship: :-)

Overall, `+` for `array` should be avoided in future, but I'll make my case stronger when I got time to investigate further: since it's already in use here, let's merge :+1:

Ocramius (approved):
LGTM (see convo @ https://github.com/laminas/laminas-servicemanager/pull/74#discussion_r559207391 )

PR #78, opened 2021-01-24
Regression in 3.6.x regarding aliased services in combination with non-shared services

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | yes

### Description

With 3.6.0, a BC break was introduced regarding non-shared services and aliases.


Ocramius (approved):
LGTM 👍

PR #93, opened 2021-07-23
Improve `ServiceManager#has()` performance

koenkivits:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

We noticed that `ServiceManager::has()` was a pretty big bottleneck on a page using a very large amount of view helpers to render a large amount of data:

image

This was on a total duration of 5.61 seconds, which means it spent 87.5% of the time checking if services exist.

The problem seemed to be caused by [the way `has()` uses dynamically created closures](https://github.com/laminas/laminas-servicemanager/blob/e76870240e368edf699c50d4bbf32607ac94145c/src/ServiceManager.php#L967-L1011). This PR refactors that code to implement the same behavior using 2 private methods, avoiding the overhead that comes with it.

We're using this patch in production now and for this specific example the performance has improved drastically:

image

The logic itself remained exactly the same, I only changed the way it was written. I hope it helps, please let me know if I need to fix or clarify anything! :)

PR #94, opened 2021-07-24
Adde benchmarks for `ServiceManager#has()`

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

In #93, the performance of `ServiceManager::has` has been improved. To have a before/after measurement, I provide this benchmark in a dedicated PR to measure current performance. After merging this to 3.7.x, rebasing #93 will give us a proper overview of the performance implications.


Ocramius:
While reviewing this locally, I discovered https://github.com/phpbench/phpbench/issues/896

I've adapted our code to work around it so far, but it is indeed a bit uglier than before. 

Will merge once CI says we're good.

Ocramius:
Thanks, @boesing!

(I took the liberty of applying CS to the `benchmarks/` directory: it was quite annoying to read)

PR #96, opened 2021-07-27
feature: replace `container-interop/container-interop`

boesing:
PoC


composer.json
```diff
@@ -44,20 +44,21 @@
         "vimeo/psalm": "^4.8"
     },
     "provide": {
-        "container-interop/container-interop-implementation": "^1.2",
```
boesing, line 54: Not sure if we have to keep this for compatibility reasons? 🤷🏼‍♂️

src/autoload.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
boesing: could be done by `psr/container` as well to get rid of `container-interop/container-interop` once and for all 🤷🏼‍♂️

composer.json
Ocramius: In theory not

src/AbstractPluginManager.php
```diff
@@ -58,17 +58,6 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
...
-        ) {
-            /**
-             * {@see \Laminas\ServiceManager\Factory\FactoryInterface} typehints
-             * against interop container and as such cannot accept non-interop
-             * psr container. Decorate it as interop.
-             */
-            $configInstanceOrParentLocator = new PsrContainerDecorator($configInstanceOrParentLocator);
-        }
         if (
             null !== $configInstanceOrParentLocator
```
Ocramius, line 68: Psalm confused here? Or did we really mess up the `$configInstanceOrParentLocator` type? :thinking:

src/Exception/ExceptionInterface.php
```diff
@@ -4,11 +4,11 @@
...
 namespace Laminas\ServiceManager\Exception;
 
-use Interop\Container\Exception\ContainerException;
+use Psr\Container\ContainerExceptionInterface;
 
 /**
  * Base exception for all Laminas\ServiceManager exceptions.
  */
-interface ExceptionInterface extends ContainerException
+interface ExceptionInterface extends ContainerExceptionInterface
```
Ocramius, line 12: This is potentially a BC break for anyone doing `} catch (\ Interop\Container\Exception\ContainerException) {`

src/Exception/ServiceNotCreatedException.php
```diff
@@ -4,15 +4,13 @@
...
 
-use Interop\Container\Exception\ContainerException;
 use RuntimeException as SplRuntimeException;
 
 /**
  * This exception is thrown when the service locator do not manage to create
  * the service (factory that has an error...)
  */
 class ServiceNotCreatedException extends SplRuntimeException implements
-    ContainerException,
```
Ocramius, line 15: This is potentially a BC break for anyone doing `} catch (\ Interop\Container\Exception\ContainerException) {`

src/Exception/ServiceNotFoundException.php
```diff
@@ -4,15 +4,15 @@
...
 use InvalidArgumentException as SplInvalidArgumentException;
+use Psr\Container\NotFoundExceptionInterface;
 
 /**
  * This exception is thrown when the service locator do not manage to find a
  * valid factory to create a service
  */
 class ServiceNotFoundException extends SplInvalidArgumentException implements
     ExceptionInterface,
-    NotFoundException
```
Ocramius, line 16: This is potentially a BC break for anyone doing `} catch (\ Interop\Container\Exception\ContainerException) {`

src/autoload.php
```diff
@@ -0,0 +1,14 @@
...
+declare(strict_types=1);
+
+use interop\container\containerinterface as InteropContainerInterface;
+use interop\container\exception\containerexception as InteropContainerException;
+use interop\container\exception\notfoundexception as InteropNotFoundException;
+use Psr\Container\ContainerExceptionInterface;
+use Psr\Container\ContainerInterface;
+use Psr\Container\NotFoundExceptionInterface;
+
+class_alias(ContainerInterface::class, InteropContainerInterface::class);
```
Ocramius, line 13: This block kinda needs to be conditional, IMO.

If `InteropContainerInterface` is not defined, we can certainly declare it ourselves, but on a system that has `container-interop/container-interop` installed, this feels like we're stepping on some other maintainer's toes.

Ocramius (changes requested):

src/Exception/ExceptionInterface.php
boesing: as it is aliased via `class_alias`, it is treated as the exact same Interface. So actually this should not be a BC break.

src/Exception/ServiceNotCreatedException.php
boesing: as it is aliased via `class_alias`, it is treated as the exact same Interface. So actually this should not be a BC break.

src/Exception/ServiceNotFoundException.php
boesing: as it is aliased via `class_alias`, it is treated as the exact same Interface. So actually this should not be a BC break.

src/autoload.php
boesing: it cannot be defined as we replace it while conflicting with older versions. thus, no need to add any conditon. and if we add psr/container v2, we still should alias the interface until we release a new major, not?

So this would only conflict if this component is used without composer.
Do you think this case should be handled?

src/AbstractPluginManager.php
boesing: Not sure if I can follow but yes, its already a Union type.

boesing: Ah, I've suppressed that for now as I dont really know how to fix that.
I think that this was an error in the baseline and now popped up due to the fact that we removed the Interop annotation.

src/Exception/ServiceNotFoundException.php
boesing: I've added integration tests to verify that this still works:
https://github.com/laminas/laminas-servicemanager/blob/1758c7e71e7d8e51d82c0e9630b89ce03fca9c68/test/ServiceManagerContainerInteropIntegrationTest.php

src/Exception/ServiceNotCreatedException.php
boesing: I've added integration tests to verify that this still works:
https://github.com/laminas/laminas-servicemanager/blob/1758c7e71e7d8e51d82c0e9630b89ce03fca9c68/test/ServiceManagerContainerInteropIntegrationTest.php

src/Exception/ExceptionInterface.php
boesing: I've added integration tests to verify that this still works:
https://github.com/laminas/laminas-servicemanager/blob/1758c7e71e7d8e51d82c0e9630b89ce03fca9c68/test/ServiceManagerContainerInteropIntegrationTest.php

boesing:
I've also added a request to `psr/container` regarding this topic.
https://github.com/php-fig/container/issues/37

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -5,7 +5,7 @@
 namespace LaminasTest\ServiceManager\AbstractFactory;
 
 use ArrayAccess;
-use Interop\Container\ContainerInterface;
+use interop\container\containerinterface;
```
weierophinney: Note the casing here. One of our phpcs rules for some reason lowercases imports and class names it cannot resolve. You'll need to add an exclusion for that rule so that rewrites like this don't happen.

weierophinney: Note the casing here. One of our phpcs rules for some reason lowercases imports and class names it cannot resolve. You'll need to add an exclusion for that rule so that rewrites like this don't happen.

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -18,12 +18,12 @@ class ReflectionBasedAbstractFactoryTest extends TestCase
 {
     use ProphecyTrait;
 
-    /** @var ContainerInterface|ObjectProphecy */
+    /** @var containerinterface|ObjectProphecy */
```
weierophinney: Notice it change here as well...

test/AbstractFactory/ReflectionBasedAbstractFactoryTest.php
```diff
@@ -18,12 +18,12 @@ class ReflectionBasedAbstractFactoryTest extends TestCase
...
     private $container;
 
     public function setUp(): void
     {
-        $this->container = $this->prophesize(ContainerInterface::class);
+        $this->container = $this->prophesize(containerinterface::class);
```
weierophinney: ... and now here.

weierophinney (changes requested):
There's a CS rule in our standard that is lowercasing class/namespace identifiers it does not recognized. I've pointed out one set, but there's more.

weierophinney:
@boesing If php-fig/container#38 gets accepted and released, how does that change this PR? I'm assuming we should wait to see what happens there?

boesing:
@weierophinney exactly. 
If `psr/container` replaces `interop` by itself, it does not need to be done here.
Would remove the `replace` stuff along with the autoload for `class_alias`, applying the changes requested in this PR and we're done.
I would prefer not to replace `interop` as we do not own it but if it gets rejected, I think we can do it (even if we shouldn't).

weierophinney:
Update for those following: see container-interop/container-interop#97 for details on the planned approach. In sum:

- That package will do a v2.0.0 release that defines its own interfaces as class aliases of the PSR-11 interfaces.
- We will adopt that in the next minor of this package, and, simultaneously, remove any references to container-interop internally. This will be a BC compatible, transparent change.
- We will document that users can now refer directly to PSR-11 interfaces in their own code once they have set the new minor version as their minimum in their own application and/or library. We will also send out messages on social media and our blog to further reach for this initiative.
- From there, we will decide whether to remove container-interop in the next minor or next major; most likely, it will be next major, as it will be a BC break for anybody upgrading from older minor versions.

Ocramius:
We kinda need to move on with this, to enable `psr/container:^2` compatibility.

@weierophinney I see no activity on `container-interop/container-interop` - do you have any leverage there?

boesing:
I recently came up with some idea.
I've stumbled upon @weierophinney [comment](https://blog.naderman.de/2014/02/17/replace-conflict-forks-explained/) on Nils Adermans Blog back in 2014 when the `replace` feature got introduced.
Matthew stated that it is inacceptable that components replace packages which are non their business.
I am +1 on this, at least we should not do that in here.

So what, if we change this PR to something which uses dependency-inversion?

### Idea
*composer.json*

```json
{
    "name": "laminas/laminas-servicemanager",
    "...": {},
    "require": {
        "php": "~7.4.0 || ~8.0.0 || ~8.1.0",
        "laminas/container-interop-replacement-implementation": "^1.0",
        "laminas/laminas-stdlib": "^3.2.1",
        "psr/container": "^1.0"
    },
    "require-dev": {
        "...": "*",
        "laminas/laminas-dummy-container-interop-replacement": "0.1.0"
    },
    "provide": {
    },
    "conflict": {
        "...": "*",
        "psr/container": "<1.2.0"
    },
    "suggest": {
        "laminas/laminas-container-interop-replacement": "Replaces the `container-interop/container-interop` package with `class_alias` to the corresponding `psr/container` interfaces rather than extending these."
    },
    "autoload": {
        "psr-4": {
            "Laminas\\ServiceManager\\": "src/"
        }
    },
    "...": {},
    "repositories": [
        {
            "type": "package",
            "package": {
                "name": "laminas/laminas-dummy-container-interop-replacement",
                "version": "0.1.0",
                "type": "metapackage",
                "replace": {
                    "container-interop/container-interop": "*"
                },
                "provide": {
                    "laminas/container-interop-replacement": "1.0.0"
                }
            }
        }
    ]
}
```

Afterwards, we create a new package called `laminas/laminas-container-interop-replacement` which actually provides the whole `class_alias`, `replace` and `provide` logic (see the `repositories` package) and upstream projects have 100% control over this.

So they can either:
1. require our replacement package `laminas/laminas-container-interop-replacement`
2. they can add their own `provide` to provide the virtual package `laminas/container-interop-replacement` in case they know what they're doing.


Afterwards, we can move forward on the plugin manager projects which actually provide implementations of the `ServiceManager` (`PluginManager`).

So in case a project only uses `laminas-servicemanager` and no `PluginManager` logic, they can migrate their code to use PSR-11 interfaces, add the `provide` section to their composer and they're done.

```json
{
    "require": {
         "laminas/laminas-servicemanager": "<version containing requirement of the virtual package>"
    },
    "provide": {
        "laminas/container-interop-replacement": "1.0.0"
    }
}
```

---

Thoughts?

Ocramius:
Not sure I fully understand: isn't `"repositories": ...` only for root packages? It would fix only our local build, but nothing for downstream consumers. 

EDIT: one major drawback is that `composer update` does not move them forward "smoothly" :|

boesing:
Yes, this approach would require upstream projects to explicitly choose their destiny.
Either, they opt-in for the new component. If they want to get rid of `container-interop` along with the laminas packages, they have to either require the `replace` package which we do provide as a dedicated component (if they do not want to modify all their code to get rid of the interop namespace) or, if they've cleaned up all their code, they can just "fake" the `provide` section in their project on their own risk.

Both will get rid of `container-interop` but requires careful handling of the upstream project to avoid any breakage.

---

Repositories is only for `laminas-service-manager` so we do not have to require that `replace` package via packagist

boesing:
These are some real-world examples:

# Project 1

Consumes only `laminas/laminas-servicemanager` for `PSR-11` implementation along with some factories for its services without actually implementing `FactoryInterface` from servicemanager (which would have required them to also `use Interop\Container\ContainerInterface;`) and thus already use `Psr\Container\ContainerInterface` like this:

```php
namespace Project1\Service\FooService;

use Psr\Container\ContainerInterface;
use Psr\Log\LoggerInterface;

final class FooServiceFactory
{
    public function __invoke(ContainerInterface $container): FooService
    {
        return new FooService($container->get(LoggerInterface::class));
    }
}
```

These can safely add this to their `composer.json` followed by an `composer update laminas/laminas-servicemanager` and the can update to the latest `laminas-servicemanager` version without actually adding an additional dependency to their project:

```json
{
    "name": "vendor/project1",
    "require": {
        "laminas/laminas-servicemanager": "^3.10.0"
    },
    "provide": {
        "laminas/container-interop-replacement": "1.0.0"
    }
}
```

# Project 2
Very similar to Project 1, does actually not use `FactoryInterface` for own factories but also requires `laminas/laminas-validator` which provides a plugin manager `ValidatorPluginManager` and is not (yet) updated to require `laminas/laminas-servicemanager` in the version which has the dependency inversion logic.

So this project could ofc. use "provide" to ignore the fact that the factories from the validator component are not compliant with the latest `servicemanager` version anymore (due to https://github.com/laminas/laminas-validator/blob/2.18.x/src/ValidatorPluginManagerFactory.php) but this would require an upstream project to take that pill as it was explicitly done by that project and thus not the problem of this component.

So until we (or the upstream project which consumes 3rd party modules which might also provide factories which are not yet compliant) do not migrate all of our components to provide compatible factories, the `replace` package would be mandatory. But it is up to the upstream project to either accept this as a dependency which does magic **or** to take full responsibility to add the `provide` to its project because it is aware that it has finished migration and thus does not need the `replace` package (anymore).

---

This is actually a different approach to the `zendframeworkbridge`. If a user does not wanted this as a part of its project, he had to opt-out. Now, we require users to opt-in. Everyone can keep the old servicemanager version, etc. in case the magic is not wanted but if one wants to get rid of container-interop without too much magic, one can either use our `replace` package or writing its own. 🤷🏼‍♂️ 


The main idea here is that upstream projects can also provide their own `class_alias` stuff in case they do not trust us regarding this magic. Or maybe there is already another packagist project which provides this `class_alias` and uses `replace`.

I guess, if we just add `replace` here, this could end-up in very weird scenarios where probably other components, which might also `replace` container-interop are being dropped because composer thinks they are also replaced by `laminas-servicemanager`.

boesing:
From a technical PoV, this provides a smooth migration path if projects are fine with adding the dedicated `replace` package. So they don't have to hassle around with all the refactorings while also keeping their deps up2date.
For projects with devs who want to have more "freedom", we allow them to handle this problem theirselves - however that might be.

So introducing a BC break how we wanted to do it in this PR but with a little bit more freedom to advanced users or projects which do not depend on `container-interop` at all.
I've thought this could be a smart idea 😅

weierophinney (approved):
Net removal of code, simplification internally, and immediate PSR-11 compatibility.

I am so thrilled by this! :heart_eyes:

PR #98, opened 2021-07-31
Refined ServiceManager factory, delegator and configuration types, to allow for easier introspection of mis-configuration at type level

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | kinda
| QA            | yes

### Description

This reduces the overall baseline by 140 errors. Some errors still have to be suppressed inline.

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -114,6 +114,8 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
     }
 
     /**
+     * Override configure() to validate service instances.
+     *
      * @param  array $config
      * @psalm-param ServiceManagerConfiguration $config
      * @return self
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -173,9 +173,8 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
             $this->setFactory($name, Factory\InvokableFactory::class);
         }
 
-        $options = $options ?? [];
         /** @psalm-suppress MixedAssignment */
-        $instance = $options === [] ? parent::get($name) : $this->build($name, $options);
+        $instance = ! $options ? parent::get($name) : $this->build($name, $options);
         $this->validate($instance);
         return $instance;
     }
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Laminas\ServiceManager;
 
 use Interop\Container\ContainerInterface;
+use Laminas\ServiceManager\Exception\ContainerModificationsNotAllowedException;
 use Laminas\ServiceManager\Exception\InvalidServiceException;
 use Psr\Container\ContainerInterface as PsrContainerInterface;
 
@@ -31,6 +32,9 @@ use const E_USER_DEPRECATED;
  *
  * The implementation extends `ServiceManager`, thus providing the same set
  * of capabilities as found in that implementation.
+ *
+ * @psalm-import-type ServiceManagerConfiguration from ServiceManager
+ * @psalm-suppress PropertyNotSetInConstructor
  */
 abstract class AbstractPluginManager extends ServiceManager implements PluginManagerInterface
 {
@@ -45,6 +49,7 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
      * An object type that the created instance must be instanced of
      *
      * @var null|string
+     * @psalm-var null|class-string
      */
     protected $instanceOf;
 
@@ -55,6 +60,7 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
      *
      * @param null|ConfigInterface|ContainerInterface|PsrContainerInterface $configInstanceOrParentLocator
      * @param array $config
+     * @psalm-param ServiceManagerConfiguration $config
      */
     public function __construct($configInstanceOrParentLocator = null, array $config = [])
     {
@@ -108,18 +114,18 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
     }
 
     /**
-     * Override configure() to validate service instances.
-     *
-     * If an instance passed in the `services` configuration is invalid for the
-     * plugin manager, this method will raise an InvalidServiceException.
-     *
-     * {@inheritDoc}
-     *
-     * @throws InvalidServiceException
+     * @param  array $config
+     * @psalm-param ServiceManagerConfiguration $config
+     * @return self
+     * @throws InvalidServiceException If an instance passed in the `services` configuration is invalid for the
+     *                                 plugin manager.
+     * @throws ContainerModificationsNotAllowedException If the allow override flag has been toggled off, and a
+     *                                                   service instanceexists for a given service.
      */
     public function configure(array $config)
     {
         if (isset($config['services'])) {
+            /** @psalm-suppress MixedAssignment */
             foreach ($config['services'] as $service) {
                 $this->validate($service);
             }
@@ -142,10 +148,8 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
     }
 
     /**
-     * {@inheritDoc}
-     *
      * @param string $name Service name of plugin to retrieve.
-     * @param null|array $options Options to use when creating the instance.
+     * @param null|array<mixed> $options Options to use when creating the instance.
      * @return mixed
      * @throws Exception\ServiceNotFoundException If the manager does not have
      *     a service definition for the instance, and the service is not
@@ -167,7 +171,9 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
             $this->setFactory($name, Factory\InvokableFactory::class);
         }
 
-        $instance = empty($options) ? parent::get($name) : $this->build($name, $options);
+        $options = $options ?? [];
+        /** @psalm-suppress MixedAssignment */
+        $instance = $options === [] ? parent::get($name) : $this->build($name, $options);
         $this->validate($instance);
         return $instance;
     }
--- a/src/Config.php
+++ b/src/Config.php
@@ -4,13 +4,9 @@ declare(strict_types=1);
 
 namespace Laminas\ServiceManager;
 
-use Laminas\Stdlib\ArrayUtils\MergeRemoveKey;
-use Laminas\Stdlib\ArrayUtils\MergeReplaceKeyInterface;
+use Laminas\Stdlib\ArrayUtils;
 
-use function array_key_exists;
 use function array_keys;
-use function is_array;
-use function is_int;
 
 /**
  * Object for defining configuration and configuring an existing service manager instance.
@@ -26,10 +22,12 @@ use function is_int;
  *
  * These features are advanced, and not typically used. If you wish to use them,
  * you will need to require the laminas-stdlib package in your application.
+ *
+ * @psalm-import-type ServiceManagerConfigurationType from ConfigInterface
  */
 class Config implements ConfigInterface
 {
-    /** @var array */
+    /** @var array<string,bool> */
     private $allowedKeys = [
         'abstract_factories' => true,
         'aliases'            => true,
@@ -42,7 +40,10 @@ class Config implements ConfigInterface
         'shared'             => true,
     ];
 
-    /** @var array */
+    /**
+     * @var array<string,array>
+     * @psalm-var ServiceManagerConfigurationType
+     */
     protected $config = [
         'abstract_factories' => [],
         'aliases'            => [],
@@ -56,7 +57,7 @@ class Config implements ConfigInterface
     ];
 
     /**
-     * @param array $config
+     * @psalm-param ServiceManagerConfigurationType $config
      */
     public function __construct(array $config = [])
     {
@@ -66,6 +67,8 @@ class Config implements ConfigInterface
                 unset($config[$key]);
             }
         }
+
+        /** @psalm-suppress ArgumentTypeCoercion */
         $this->config = $this->merge($this->config, $config);
     }
 
@@ -86,32 +89,14 @@ class Config implements ConfigInterface
     }
 
     /**
-     * Copy paste from https://github.com/laminas/laminas-stdlib/commit/26fcc32a358aa08de35625736095cb2fdaced090
-     * to keep compatibility with previous version
-     *
-     * @link https://github.com/zendframework/zend-servicemanager/pull/68
+     * @psalm-param ServiceManagerConfigurationType $a
+     * @psalm-param ServiceManagerConfigurationType $b
+     * @psalm-return ServiceManagerConfigurationType
+     * @psalm-suppress MixedReturnTypeCoercion
      */
-    private function merge(array $a, array $b): array
+    private function merge(array $a, array $b)
     {
-        foreach ($b as $key => $value) {
-            if ($value instanceof MergeReplaceKeyInterface) {
-                $a[$key] = $value->getData();
-            } elseif (isset($a[$key]) || array_key_exists($key, $a)) {
-                if ($value instanceof MergeRemoveKey) {
-                    unset($a[$key]);
-                } elseif (is_int($key)) {
-                    $a[] = $value;
-                } elseif (is_array($value) && is_array($a[$key])) {
-                    $a[$key] = $this->merge($a[$key], $value);
-                } else {
-                    $a[$key] = $value;
-                }
-            } else {
-                if (! $value instanceof MergeRemoveKey) {
-                    $a[$key] = $value;
-                }
-            }
-        }
-        return $a;
+        /** @psalm-suppress MixedReturnTypeCoercion */
+        return ArrayUtils::merge($a, $b);
     }
 }
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -20,14 +20,14 @@ use Interop\Container\ContainerInterface;
  *      array<
  *          array-key,
  *          (class-string<Factory\DelegatorFactoryInterface>|Factory\DelegatorFactoryInterface)
- *          |callable(ContainerInterface,string,callable():object,array<mixed>):object
+ *          |callable(ContainerInterface,string,callable():object,array<mixed>|null):object
  *          |(class-string<DelegatorFactoryInterface>|DelegatorFactoryInterface)
  *      >
  * >
  * @psalm-type FactoriesConfigurationType = array<
  *      string,
  *      (class-string<Factory\FactoryInterface>|Factory\FactoryInterface)
- *      |callable(ContainerInterface,string,array<mixed>)
+ *      |callable(ContainerInterface,string,array<mixed>|null)
  * >
  * @psalm-type InitializersConfigurationType = array<
  *      array-key,
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -4,6 +4,56 @@ declare(strict_types=1);
 
 namespace Laminas\ServiceManager;
 
+use ArrayAccess;
+use Interop\Container\ContainerInterface;
+
+/**
+ * @see ContainerInterface
+ * @see ArrayAccess
+ *
+ * @psalm-type AbstractFactoriesConfigurationType = array<
+ *      array-key,
+ *      (class-string<Factory\AbstractFactoryInterface>|Factory\AbstractFactoryInterface)
+ * >
+ * @psalm-type DelegatorsConfigurationType = array<
+ *      string,
+ *      array<
+ *          array-key,
+ *          (class-string<Factory\DelegatorFactoryInterface>|Factory\DelegatorFactoryInterface)
+ *          |class-string
+ *          |callable(ContainerInterface,string,callable():object,array<mixed>):object
+ *          |(class-string<DelegatorFactoryInterface>|DelegatorFactoryInterface)
+ *      >
+ * >
+ * @psalm-type FactoriesConfigurationType = array<
+ *      string,
+ *      (class-string<Factory\FactoryInterface>|Factory\FactoryInterface)
+ *      |class-string
+ *      |callable(ContainerInterface,string,array<mixed>)
+ * >
+ * @psalm-type InitializersConfigurationType = array<
+ *      array-key,
+ *      (class-string<Initializer\InitializerInterface>|Initializer\InitializerInterface)
+ *      |callable(ContainerInterface,object):void
+ * >
+ * @psalm-type LazyServicesConfigurationType = array{
+ *      class_map?:array<string,class-string>,
+ *      proxies_namespace?:non-empty-string,
+ *      proxies_target_dir?:non-empty-string,
+ *      write_proxy_files?:bool
+ * }
+ * @psalm-type ServiceManagerConfigurationType = array{
+ *     abstract_factories?: AbstractFactoriesConfigurationType,
+ *     aliases?: array<string,string>,
+ *     delegators?: DelegatorsConfigurationType,
+ *     factories?: FactoriesConfigurationType,
+ *     initializers?: InitializersConfigurationType,
+ *     invokables?: array<string,string>,
+ *     lazy_services?: LazyServicesConfigurationType,
+ *     services?:array<string,object|array>,
+ *     shared?:array<string,bool>
+ * }
+ */
 interface ConfigInterface
 {
     /**
@@ -37,6 +87,7 @@ interface ConfigInterface
      * a service manager or plugin manager, or pass to its `withConfig()` method.
      *
      * @return array
+     * @psalm-return ServiceManagerConfigurationType
      */
     public function toArray();
 }
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -50,7 +50,7 @@ use Interop\Container\ContainerInterface;
  *     initializers?: InitializersConfigurationType,
  *     invokables?: array<string,string>,
  *     lazy_services?: LazyServicesConfigurationType,
- *     services?:array<string,object|array>,
+ *     services?: array<string,object|array>,
  *     shared?:array<string,bool>
  * }
  */
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -20,7 +20,6 @@ use Interop\Container\ContainerInterface;
  *      array<
  *          array-key,
  *          (class-string<Factory\DelegatorFactoryInterface>|Factory\DelegatorFactoryInterface)
- *          |class-string
  *          |callable(ContainerInterface,string,callable():object,array<mixed>):object
  *          |(class-string<DelegatorFactoryInterface>|DelegatorFactoryInterface)
  *      >
@@ -28,7 +27,6 @@ use Interop\Container\ContainerInterface;
  * @psalm-type FactoriesConfigurationType = array<
  *      string,
  *      (class-string<Factory\FactoryInterface>|Factory\FactoryInterface)
- *      |class-string
  *      |callable(ContainerInterface,string,array<mixed>)
  * >
  * @psalm-type InitializersConfigurationType = array<
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -21,7 +21,6 @@ use Interop\Container\ContainerInterface;
  *          array-key,
  *          (class-string<Factory\DelegatorFactoryInterface>|Factory\DelegatorFactoryInterface)
  *          |callable(ContainerInterface,string,callable():object,array<mixed>|null):object
- *          |(class-string<DelegatorFactoryInterface>|DelegatorFactoryInterface)
  *      >
  * >
  * @psalm-type FactoriesConfigurationType = array<
--- a/src/Factory/DelegatorFactoryInterface.php
+++ b/src/Factory/DelegatorFactoryInterface.php
@@ -23,6 +23,7 @@ interface DelegatorFactoryInterface
      * A factory that creates delegates of a given service
      *
      * @param  string             $name
+     * @psalm-param callable():mixed $callback
      * @param  null|array         $options
      * @return object
      * @throws ServiceNotFoundException If unable to resolve the service.
--- a/src/PluginManagerInterface.php
+++ b/src/PluginManagerInterface.php
@@ -17,11 +17,12 @@ interface PluginManagerInterface extends ServiceLocatorInterface
     /**
      * Validate an instance
      *
-     * @param  object $instance
+     * @param  mixed $instance
      * @return void
      * @throws InvalidServiceException If created instance does not respect the
      *     constraint on type imposed by the plugin manager.
      * @throws ContainerException If any other error occurs.
+     * @psalm-assert TPlugin $instance
      */
     public function validate($instance);
 }
--- a/src/PluginManagerInterface.php
+++ b/src/PluginManagerInterface.php
@@ -22,7 +22,6 @@ interface PluginManagerInterface extends ServiceLocatorInterface
      * @throws InvalidServiceException If created instance does not respect the
      *     constraint on type imposed by the plugin manager.
      * @throws ContainerException If any other error occurs.
-     * @psalm-assert TPlugin $instance
      */
     public function validate($instance);
 }
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -118,7 +118,7 @@ class ServiceManager implements ServiceLocatorInterface
     /**
      * A list of already loaded services (this act as a local cache)
      *
-     * @var array<string,mixed>
+     * @var array<string,array|object>
      */
     protected $services = [];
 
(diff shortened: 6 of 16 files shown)


src/AbstractPluginManager.php
```diff
@@ -31,6 +33,15 @@
  *
  * The implementation extends `ServiceManager`, thus providing the same set
  * of capabilities as found in that implementation.
+ *
+ * @see ServiceLocatorInterface
```
boesing: I use `@see` because the codesniffers out there as well as phpstorm are not able to detect the usage from within `@psalm-import-type`.

src/Config.php
```diff
@@ -86,32 +89,14 @@ public function toArray()
...
-                }
-            } else {
-                if (! $value instanceof MergeRemoveKey) {
-                    $a[$key] = $value;
-                }
-            }
-        }
-        return $a;
+        /** @psalm-suppress MixedReturnTypeCoercion */
+        return ArrayUtils::merge($a, $b);
```
boesing, line 100: I removed the inlined code here as `laminas-stdlib` is a dependency within `ServiceManager` to use `merge` for the `lazy_services`.

src/ServiceManager.php
```diff
@@ -271,40 +291,8 @@ public function getAllowOverride()
     }
 
     /**
-     * Configure the service manager
-     *
-     * Valid top keys are:
-     *
-     * - services: service name => service instance pairs
```
boesing, line 275: these details are obsolete as we are now providing a proper psalm type.
so imho this can be deleted, right?

src/ServiceManager.php
```diff
@@ -415,8 +404,11 @@ public function setInvokableClass($name, $class = null)
      * Specify a factory for a given service name.
      *
      * @param string $name Service name
-     * @param string|callable|Factory\FactoryInterface $factory Factory to which
-     *     to map.
+     * @param string|callable|Factory\FactoryInterface $factory  Factory to which to map.
+     * phpcs:disable Generic.Files.LineLength.TooLong
+     * @psalm-param class-string<Factory\FactoryInterface>|callable(ContainerInterface,string,ServiceOptionsType|null)|Factory\FactoryInterface $factory
```
boesing: it seems that `@psalm-param` cannot properly detect multiline declaration and thus, I had to use `phpcs:disable` to disable the line length check. 
Screw you, line length check!

At least I was not able to do it here, other methods worked obviously. Not sure what the reason is but TBH, I don't really care and I don't want to spend more time on this just to make the codestyle happy.

src/AbstractPluginManager.php
```diff
@@ -31,6 +33,15 @@
...
+ *
+ * @template TPlugin of object
```
boesing: So finally we can start providing generics for plugin managers

src/PluginManagerInterface.php
```diff
@@ -11,6 +11,9 @@
  * Interface for a plugin manager
  *
  * A plugin manager is a specialized service locator used to create homogeneous objects
+ *
+ * @template TPlugin of object
+ * @method build($name, ?array $options = null): TPlugin
```
boesing: TBH, I did not verify if that really works.

src/ServiceManager.php
```diff
@@ -78,29 +86,39 @@ class ServiceManager implements ServiceLocatorInterface
...
 
-    /** @var null|Proxy\LazyServiceFactory */
+    /** @var null|LazyServiceFactory */
     private $lazyServicesDelegator;
 
     /**
      * A list of already loaded services (this act as a local cache)
      *
-     * @var array
+     * @var array<string,mixed>
```
boesing: I've seen somewhere that only `array|object` is allowed for services. Maybe something to consider here?

boesing: https://github.com/laminas/laminas-servicemanager/pull/98/files#diff-c88e8922673ece50d04aef21f4dc95b87bd465868eb073480467f87e626e3b0aR478

src/ConfigInterface.php
```diff
@@ -4,6 +4,56 @@
...
+ * }
+ * @psalm-type ServiceManagerConfigurationType = array{
+ *     abstract_factories?: AbstractFactoriesConfigurationType,
+ *     aliases?: array<string,string>,
+ *     delegators?: DelegatorsConfigurationType,
+ *     factories?: FactoriesConfigurationType,
+ *     initializers?: InitializersConfigurationType,
+ *     invokables?: array<string,string>,
+ *     lazy_services?: LazyServicesConfigurationType,
+ *     services?:array<string,object>&array{config?:ArrayAccess<string,mixed>|array<string,mixed>},
```
boesing: https://github.com/laminas/laminas-servicemanager/pull/98/files#diff-c88e8922673ece50d04aef21f4dc95b87bd465868eb073480467f87e626e3b0aR478

This allows `array|object`.
Yes, I know that this would not only allow `config` but also other arrays.
As @Ocramius pointed out in #99, this might be a BC break (even tho, we still accept arrays), so should I change this to `array<string,object|array>` instead?

src/AbstractPluginManager.php
```diff
@@ -108,14 +118,13 @@ public function __construct($configInstanceOrParentLocator = null, array $config
     }
 
     /**
-     * Override configure() to validate service instances.
```
weierophinney, line 111: This should still be in the docblock, as it gives a description of what we are doing here while overriding the parent.

src/Config.php
weierophinney: :+1:

src/ConfigInterface.php
weierophinney: yes.

weierophinney: yes.

src/ServiceManager.php
weierophinney: Yeah - only array or object are allowed as return types.

src/ServiceManager.php
weierophinney: I would say "yes', because:

- We have end-user documentation detailing what is accepted for each.
- The Psalm types give IDEs the information necessary to provide hinting to developers.

src/ServiceManager.php
```diff
@@ -412,8 +403,11 @@ public function setInvokableClass($name, $class = null)
      * Specify a factory for a given service name.
      *
      * @param string $name Service name
-     * @param string|callable|Factory\FactoryInterface $factory Factory to which
-     *     to map.
+     * @param string|callable|Factory\FactoryInterface $factory  Factory to which to map.
+     * phpcs:disable Generic.Files.LineLength.TooLong
+     * @psalm-param class-string<Factory\FactoryInterface>|callable(ContainerInterface,string,array<mixed>|null)|Factory\FactoryInterface $factory
+     * phpcs:enable Generic.Files.LineLength.TooLong
```
weierophinney, line 409:
```suggestion
     * @psalm-param class-string<Factory\FactoryInterface>
     *     |callable(ContainerInterface,string,array<mixed>|null)
     *     |Factory\FactoryInterface $factory
```

( Matches what you did on lines 453-454, and eliminates the need for the phpcs directives)

src/ServiceManager.php
```diff
@@ -578,55 +581,38 @@ private function createDelegatorFromName($name, ?array $options = null)
...
-                    ));
-                }
-
-                throw new ServiceNotCreatedException(sprintf(
-                    'A non-callable delegator, "%s", was provided; expected a callable or instance of "%s"',
-                    is_object($delegatorFactory) ? get_class($delegatorFactory) : gettype($delegatorFactory),
-                    DelegatorFactoryInterface::class
-                ));
-            }
+            $this->assertCallableDelegatorFactory($delegatorFactory);
```
weierophinney, line 588: :+1:

Love the method extraction.

src/ServiceManager.php
```diff
@@ -968,4 +958,33 @@ private function abstractFactoryCanCreate(string $name): bool
...
+                ));
+            }
+
+            throw new ServiceNotCreatedException(sprintf(
+                'A non-callable delegator, "%s", was provided; expected a callable or instance of "%s"',
+                is_object($delegatorFactory) ? get_class($delegatorFactory) : gettype($delegatorFactory),
+                DelegatorFactoryInterface::class
+            ));
+        }
+    }
```
weierophinney:
```suggestion
        if ($delegatorFactory instanceof Factory\DelegatorFactoryInterface
            || is_callable($delegatorFactory)
        ) {
            return;
        }

        if (is_string($delegatorFactory)) {
            throw new ServiceNotCreatedException(sprintf(
                'An invalid delegator factory was registered; resolved to class or function "%s"'
                . ' which does not exist; please provide a valid function name or class name resolving'
                . ' to an implementation of %s',
                $delegatorFactory,
                DelegatorFactoryInterface::class
            ));
        }

      throw new ServiceNotCreatedException(sprintf(
            'A non-callable delegator, "%s", was provided; expected a callable or instance of "%s"',
            is_object($delegatorFactory) ? get_class($delegatorFactory) : gettype($delegatorFactory),
            DelegatorFactoryInterface::class
        ));
    }
```

Basically, return early as soon as we know it's valid, and otherwise just vary the exception based on whether or not it was passed as a string.

test/AbstractPluginManagerTest.php
```diff
@@ -365,7 +365,7 @@ public function testValidateWillFallBackToValidatePluginWhenDefinedAndEmitDeprec
         $pluginManager->assertion = $assertion;
 
         $errorHandlerCalled = false;
-        set_error_handler(function ($errno, $errmsg) use (&$errorHandlerCalled) {
+        set_error_handler(function (int $errno, string $errmsg) use (&$errorHandlerCalled) {
```
weierophinney:
```suggestion
        set_error_handler(function (int $errno, string $errmsg) use (&$errorHandlerCalled): void {
```

Add the return type.

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -442,8 +448,9 @@ public function testCanConfigureAllServiceTypes(): void
             ],
             'delegators'         => [
                 stdClass::class => [
-                    function ($container, $name, $callback) {
-                        $instance      = $callback();
+                    function (ContainerInterface $container, string $name, callable $callback) {
```
weierophinney:
```suggestion
                    function (ContainerInterface $container, string $name, callable $callback): object {
```

test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -899,7 +910,8 @@ public function testAllowOverrideFlagIsMutable(ServiceManager $container): void
     public function testCanRetrieveParentContainerViaGetServiceLocatorWithDeprecationNotice(): void
     {
         $container = $this->createContainer();
-        set_error_handler(function ($errno, $errstr) {
+        /** @psalm-suppress InvalidArgument */
+        set_error_handler(function (int $errno) {
```
weierophinney:
```suggestion
        set_error_handler(function (int $errno): void {
```

weierophinney (dismissed):
I think this is a GREAT addition! I've flagged a few nitpicks, which are primarily to close a few more Psalm errors as well as simplify some of the methods by promoting early returns.

Otherwise, :ship:

src/ServiceManager.php
boesing: ~but then we have to bc break `ServiceManager::doCreate` as it allows to return `mixed`~

its private. I'll add native type-hint which might break stuff upstream as it is definitely possible to return any type from a factory even tho it states one has to return objects.

src/ServiceManager.php
boesing: When I do this locally, I get a psalm error.

```
ERROR: PropertyTypeCoercion - src/ServiceManager.php:421:9 - $this->factories expects 'array<string, Laminas\ServiceManager\Factory\FactoryInterface|callable(Interop\Container\ContainerInterface, string, array<array-key, mixed>)|class-string>',  parent type 'non-empty-array<string, Laminas\ServiceManager\Factory\FactoryInterface|callable|callable(Interop\Container\ContainerInterface, string, array<array-key, mixed>)|string>' provided (see https://psalm.dev/198)
        $this->factories[$name] = $factory;
```

Somehow there is an additional `callable` in there. I guess its due to the `null` type-hint

boesing:
After some talks to @Ocramius, we are thinking about extracting the `@template`-Annotation into a dedicated patch to avoid upstream analysis errors.

The dedicated patch would target v4.0 which does not have any ETA and which would require a huge amount of work in a majority of our components. Maybe this is something for the hacktoberfest?

src/AbstractPluginManager.php
```diff
@@ -31,6 +33,11 @@
  *
  * The implementation extends `ServiceManager`, thus providing the same set
  * of capabilities as found in that implementation.
+ *
+ * @template TPlugin of object
```
Ocramius: As discussed in chat, probably worth dropping this from this specific patch

src/Config.php
```diff
@@ -86,32 +89,14 @@ public function toArray()
...
 
     /**
-     * Copy paste from https://github.com/laminas/laminas-stdlib/commit/26fcc32a358aa08de35625736095cb2fdaced090
-     * to keep compatibility with previous version
-     *
-     * @link https://github.com/zendframework/zend-servicemanager/pull/68
+     * @psalm-param ServiceManagerConfigurationType $a
+     * @psalm-param ServiceManagerConfigurationType $b
+     * @psalm-return ServiceManagerConfigurationType
+     * @psalm-suppress MixedReturnTypeCoercion
```
Ocramius, line 95: This suppression should only be in either the body of the method, or the docblock, but not both

src/ConfigInterface.php
```diff
@@ -4,6 +4,56 @@
...
+ * @psalm-type AbstractFactoriesConfigurationType = array<
+ *      array-key,
+ *      (class-string<Factory\AbstractFactoryInterface>|Factory\AbstractFactoryInterface)
+ * >
+ * @psalm-type DelegatorsConfigurationType = array<
+ *      string,
+ *      array<
+ *          array-key,
+ *          (class-string<Factory\DelegatorFactoryInterface>|Factory\DelegatorFactoryInterface)
+ *          |class-string
```
Ocramius: Why is any `class-string` allowed?

src/ConfigInterface.php
```diff
@@ -4,6 +4,56 @@
...
+ *          |callable(ContainerInterface,string,callable():object,array<mixed>):object
+ *          |(class-string<DelegatorFactoryInterface>|DelegatorFactoryInterface)
+ *      >
+ * >
+ * @psalm-type FactoriesConfigurationType = array<
+ *      string,
+ *      (class-string<Factory\FactoryInterface>|Factory\FactoryInterface)
+ *      |class-string
```
Ocramius: Why is any `class-string` allowed?

src/Factory/DelegatorFactoryInterface.php
```diff
@@ -23,6 +23,7 @@ interface DelegatorFactoryInterface
      * A factory that creates delegates of a given service
      *
      * @param  string             $name
+     * @psalm-param callable():mixed $callback
      * @param  null|array         $options
      * @return object
```
Ocramius, line 28: If this is supposed to return `object`, then `$callback` cannot return `mixed`.

I feel like the `DelegatorFactoryInterface` may benefit from a templated type, in future (not here, due to BC)

src/PluginManagerInterface.php
Ocramius: This method really shouldn't be declared here, since otherwise we require child implementations to include a `build()` method, or face scenarios where the child class does not implement it, yet psalm considers it existing => runtime crash.

Let's drop it.

src/ConfigInterface.php
```diff
@@ -4,6 +4,56 @@
...
+ *      |callable(ContainerInterface,string,array<mixed>)
```
Ocramius: I think `array<mixed>|null` is needed, to allow for factories to deal with the "no config provided" scenario?

src/ConfigInterface.php
```diff
@@ -4,6 +4,56 @@
...
+ *      array-key,
+ *      (class-string<Factory\AbstractFactoryInterface>|Factory\AbstractFactoryInterface)
+ * >
+ * @psalm-type DelegatorsConfigurationType = array<
+ *      string,
+ *      array<
+ *          array-key,
+ *          (class-string<Factory\DelegatorFactoryInterface>|Factory\DelegatorFactoryInterface)
+ *          |class-string
+ *          |callable(ContainerInterface,string,callable():object,array<mixed>):object
```
Ocramius: I think `array<mixed>|null` is needed, to allow for delegators to deal with the "no config provided" scenario?

src/ServiceManager.php
```diff
@@ -455,6 +450,8 @@ public function addAbstractFactory($factory)
      * @param string $name Service name
      * @param string|callable|Factory\DelegatorFactoryInterface $factory Delegator
      *     factory to assign.
+     * @psalm-param class-string<Factory\DelegatorFactoryInterface>
+     *     |callable(ContainerInterface,string,callable,array<mixed>) $factory
```
Ocramius: I think `array<mixed>|null` is needed

src/ServiceManager.php
```diff
@@ -412,8 +403,11 @@ public function setInvokableClass($name, $class = null)
      * Specify a factory for a given service name.
      *
      * @param string $name Service name
-     * @param string|callable|Factory\FactoryInterface $factory Factory to which
-     *     to map.
+     * @param string|callable|Factory\FactoryInterface $factory  Factory to which to map.
+     * phpcs:disable Generic.Files.LineLength.TooLong
+     * @psalm-param class-string<Factory\FactoryInterface>|callable(ContainerInterface,string,array<mixed>|null)|Factory\FactoryInterface $factory
```
Ocramius, line 408: Should we declare some of the more complex union types such as `class-string<Factory\FactoryInterface>|callable(ContainerInterface,string,array<mixed>|null)|Factory\FactoryInterface` as `FactoryType` in psalm, or such?

src/ServiceManager.php
```diff
@@ -507,16 +507,20 @@ public function setShared($name, $flag)
     /**
      * Instantiate initializers for to avoid checks during service construction.
      *
-     * @param string[]|Initializer\InitializerInterface[]|callable[] $initializers
+     * @psalm-param array<
+     *     array-key,
+     *     class-string<Initializer\InitializerInterface>|Initializer\InitializerInterface|callable
```
Ocramius: The `callable` type here should be typed better: removes the need for declaring  `@psalm-var` below

src/ServiceManager.php
```diff
@@ -507,16 +507,20 @@ public function setShared($name, $flag)
...
-    private function resolveInitializers(array $initializers)
+    private function resolveInitializers(array $initializers): void
     {
         foreach ($initializers as $initializer) {
             if (is_string($initializer) && class_exists($initializer)) {
                 $initializer = new $initializer();
             }
 
             if (is_callable($initializer)) {
+                /** @psalm-var callable(ContainerInterface,object):void */
```
Ocramius: Becomes unnecessary if we type the above  `callable` a bit better

src/ServiceManager.php
```diff
@@ -543,9 +546,11 @@ private function getFactory($name)
         }
 
         if (is_callable($factory)) {
+            /** @psalm-var callable(ContainerInterface,string,array<mixed>|null):object $factory */
```
Ocramius, line 545: shouldn't be needed? Perhaps this is because of the missing `|null` in above factory definitions (for the options parameter)?

src/ServiceManager.php
```diff
@@ -578,55 +581,37 @@ private function createDelegatorFromName($name, ?array $options = null)
...
-                throw new ServiceNotCreatedException(sprintf(
-                    'A non-callable delegator, "%s", was provided; expected a callable or instance of "%s"',
-                    is_object($delegatorFactory) ? get_class($delegatorFactory) : gettype($delegatorFactory),
-                    DelegatorFactoryInterface::class
-                ));
-            }
+            $this->assertCallableDelegatorFactory($delegatorFactory);
 
+            /**
+             * @psalm-var callable(ContainerInterface,string,callable,array<mixed>|null):object $delegatorFactory
```
Ocramius: Should not be needed, especially after  `psalm-assert`

src/ServiceManager.php
```diff
@@ -578,55 +581,37 @@ private function createDelegatorFromName($name, ?array $options = null)
...
+             */
             $this->delegators[$name][$index] = $delegatorFactory;
 
             $creationCallback = function () use ($delegatorFactory, $name, $creationCallback, $options) {
```
Ocramius: I think this closure should be `static` - we're probably leaking GC (and memory, obviously) here...

src/ServiceManager.php
```diff
@@ -578,55 +581,37 @@ private function createDelegatorFromName($name, ?array $options = null)
...
      *
-     * @param  string     $resolvedName
-     * @param  null|array $options
-     * @return mixed
      * @throws ServiceNotFoundException If unable to resolve the service.
      * @throws ServiceNotCreatedException If an exception is raised when creating a service.
      * @throws ContainerException If any other error occurs.
      */
-    private function doCreate($resolvedName, ?array $options = null)
+    private function doCreate(string $resolvedName, ?array $options = null): object
```
Ocramius: This will break with `array` services, due to the newly declared `: object` return value

test/ConfigTest.php
```diff
@@ -106,6 +106,7 @@ public function testPassesKnownServiceConfigKeysToServiceManagerWithConfigMethod
         $services = $this->prophesize(ServiceManager::class);
         $services->configure($expected)->willReturn('CALLED');
 
+        /** @psalm-suppress InvalidArgument */
```
Ocramius: Let's add a comment here - ideally, test should be adapted, but it's fine to show that the system keeps working as it did before

test/ContainerTest.php
```diff
@@ -6,15 +6,22 @@
...
+ *
+ * @psalm-import-type ServiceManagerConfigurationType from ConfigInterface
+ */
 class ContainerTest extends AbstractMezzioContainerConfigTest
 {
     use SharedTestTrait;
 
     protected function createContainer(array $config): ContainerInterface
     {
+        /** @psalm-suppress MixedArgumentTypeCoercion */
```
Ocramius, line 18: The base `createContainer` method should use the `ServiceManagerConfigurationType`, but I see it's imported here, and never used

test/ServiceManagerTest.php
```diff
@@ -475,7 +487,6 @@ public function testConfigureMultipleTimesAvoidsDuplicates(): void
         ];
 
         $serviceManager = new ServiceManager($config);
-        $serviceManager->configure($config);
```
Ocramius: This test is explicitly called `testConfigureMultipleTimesAvoidsDuplicates` - it's designed to do the configuration multiple times, so this call should stay

Ocramius (changes requested):

src/Config.php
boesing: it actually has to be there. if you check the psalm config, I've added "report unused psalm suppress".

src/ConfigInterface.php
boesing: Good catch. Its not.

src/Factory/DelegatorFactoryInterface.php
boesing: Since `ContainerInterface#get` can return `mixed`, this actually has to be `mixed` :-(

But yes, having a templated type would help for sure but until then, I've thought that `mixed` would be fine until the return type is object 🤷🏼‍♂️

src/PluginManagerInterface.php
boesing: The method is declared in the `ServiceLocatorInterface` which is extended from this interface.
I just wanted to tighten the return type of the `build` method but I forgot to add `get`-method as well.

Will do so in the patch for the next major. until then, I drop it.

src/ConfigInterface.php
boesing: good catch

src/ServiceManager.php
boesing: TBH; I am happy that psalm is green. I've now spent around 1 hour to extract some types into psalm types without success.

Would prefer doing so in another patch.

Ocramius: Yeah, it is an isomorphic change anyway 👍

src/ServiceManager.php
boesing: Nah, the problem is, that a few lines above, `$factory` is instantiated due to `class_exists` and thus, it *probably* could be any object implementing `__invoke` (which is actually true).

And thus, psalm cannot properly infer that the instantiated object is `__callable` while following the rules of the `callable` annotation.

---

Took me a bit until I understood this: https://github.com/vimeo/psalm/issues/5384

src/ServiceManager.php
boesing: `array` services are not retrievable via `doCreate` as this method only uses `factories` or `abstract_factories` which both must return `object` as per definition.

`array` is "usually" only `ServiceManager#setService('config', $configArray);` or somehow else but it actually is not possible to create an array via a factory.

anyways, I've converted this to a soft type-hint

test/ConfigTest.php
boesing: I've added a comment. I already forgot why I've added this initially TBH.

test/ContainerTest.php
boesing: I can either suppress this error or `MoreSpecificImplementedParamType` due to the implementation details here: 
https://github.com/laminas/laminas-container-config-test/blob/6a7c3f3fed29ed844cdc22b1d911cb2ef745873e/src/AbstractContainerTest.php#L18

I don't know what would be better tho.

src/AbstractPluginManager.php
boesing: Removed from this patch.

src/AbstractPluginManager.php
```diff
@@ -110,16 +116,18 @@ public function __construct($configInstanceOrParentLocator = null, array $config
...
+     * @return self
+     * @throws InvalidServiceException If an instance passed in the `services` configuration is invalid for the
+     *                                 plugin manager.
+     * @throws ContainerModificationsNotAllowedException If the allow override flag has been toggled off, and a
+     *                                                   service instanceexists for a given service.
      */
     public function configure(array $config)
     {
         if (isset($config['services'])) {
+            /** @psalm-suppress MixedAssignment */
```
Ocramius, line 130: This should be fixable: `@psalm-param ServiceManagerConfiguration|array{services: ServiceManagerConfiguration} $config`

src/AbstractPluginManager.php
```diff
@@ -167,7 +173,9 @@ public function get($name, ?array $options = null)
             $this->setFactory($name, Factory\InvokableFactory::class);
         }
 
-        $instance = empty($options) ? parent::get($name) : $this->build($name, $options);
+        $options = $options ?? [];
```
Ocramius: Instead of creating a new array (expensive), let's check if it is `null`, and call `build()` if it is **not** null? I would also expect an explicit call to `build()` even if an empty array is given...

Ocramius: Keeping `empty()` could also be a sufficient solution for now

src/Config.php
```diff
@@ -26,10 +22,12 @@
  *
  * These features are advanced, and not typically used. If you wish to use them,
  * you will need to require the laminas-stdlib package in your application.
+ *
+ * @psalm-import-type ServiceManagerConfigurationType from ConfigInterface
  */
 class Config implements ConfigInterface
 {
-    /** @var array */
+    /** @var array<string,bool> */
```
Ocramius, line 30: Not important, but I noticed that your types have no spacing in it 😁

src/Config.php
```diff
@@ -66,6 +67,8 @@ public function __construct(array $config = [])
                 unset($config[$key]);
             }
         }
+
+        /** @psalm-suppress ArgumentTypeCoercion */
```
Ocramius, line 71: This coercion shouldn't apply anymore: `merge()` has the correct types? 🤔

src/Factory/DelegatorFactoryInterface.php
Ocramius: Yeah, but the commented line is about `@return object` (wrong)

src/PluginManagerInterface.php
```diff
@@ -17,11 +17,12 @@ interface PluginManagerInterface extends ServiceLocatorInterface
     /**
      * Validate an instance
      *
-     * @param  object $instance
+     * @param  mixed $instance
      * @return void
      * @throws InvalidServiceException If created instance does not respect the
      *     constraint on type imposed by the plugin manager.
      * @throws ContainerException If any other error occurs.
+     * @psalm-assert TPlugin $instance
```
Ocramius: Leftover?

src/ServiceManager.php
```diff
@@ -563,70 +564,58 @@ private function getFactory($name)
...
+            $this->assertCallableDelegatorFactory($delegatorFactory);
 
             $this->delegators[$name][$index] = $delegatorFactory;
 
-            $creationCallback = function () use ($delegatorFactory, $name, $creationCallback, $options) {
-                return $delegatorFactory($this->creationContext, $name, $creationCallback, $options);
-            };
+            $creationCallback =
+                /** @return object */
+                static function () use (
```
Ocramius, line 594: Making this a static function is a huge improvement!

test/ContainerTest.php
Ocramius: This was more a note on the fact that the imported type in this file is not used. The class-level docblock can probably be dropped

Ocramius (changes requested):

src/AbstractPluginManager.php
boesing: That would be a false type as `services` does not contain another `ServiceManagerConfiguration`. 
It contains `array<string,object|array>` which is properly annotated but somehow psalm refuses to infer it from the parameter type.

```php
/**
 * @psalm-type ServiceManagerConfigurationType = array{
 *     services?: array<string,object|array>
 * }
 */
```

src/Config.php
boesing: why would they need spacings? is that another coding style thing? 
The more spaces, the merrier line length issues on complex types 😩

src/Factory/DelegatorFactoryInterface.php
boesing: Not sure if I can follow.

Do you want me to change existing documentation that we end up having bc breaks? as of now, due to the lack of typing, callable factories can return mixed. When I enforce `object` now, this will be pain probably. Thats why I think that might be a BC break but I am not 100% sure tho.

Or do you want me to open up the return value of the delegator to be mixed?

test/ContainerTest.php
boesing: Ah yes, I've added it initially but then removed it again. Will drop that.

src/AbstractPluginManager.php
boesing: I've added `!$options` - that would avoid calling `empty` and does not require a new array.
Would that be sufficient?

src/Config.php
boesing: `ArrayUtils::merge` returns `array` and thus, psalm loses track.

Not sure if something like this would help psalm to keep track:

```
/**
 * @template TArray1 of array
 * @template TArray2 of array
 * @param TArray1 $a
 * @param TArray2 $b
 * @return TArray1&TArray2
 */
```

But this wont properly work for nested arrays (as they are here). There is absolutely no way but to trust that `ArrayUtils::merge` will return the same types.

src/Factory/DelegatorFactoryInterface.php
boesing: I think I am fine with both as I don't have that much experience with adding psalm types on such a widely used component. As always, I trust your experience. :-)

src/AbstractPluginManager.php
Ocramius: Ah, I misread it - thought it was a different nesting level

src/Config.php
Ocramius: Types are not just an afterthought - should be as readable as other stuff 😛

src/Config.php
Ocramius: Yes, but you are not calling `ArrayUtils::merge()`, but rather `$this->merge()`, which is well typed, and has its own suppressions

Ocramius: I'll leave this alone for now - I think it's not worth wasting effort on it after all the work you've already put in this, and it's in the internals anyway :)

test/ServiceManagerTest.php
```diff
@@ -20,16 +21,23 @@
 use function get_class;
 
 /**
+ * @see ConfigInterface
```
Ocramius, line 24: I really like how you used `@see` to import types that are otherwise elided by the CS tooling :D

Ocramius (approved):

Ocramius:
🚢 

will also release ASAP - it's a huge improvement :)

src/Config.php
boesing: I've added `reportUnusedPsalmSuppress` in the config and thats not reported so something might still be odd :-/

PR #99, opened 2021-08-02
Explicitly declare that factories support `array<mixed>` for construct-time options, instead of just declaring `array`

fezfez:
see https://github.com/Slamdunk/phpstan-laminas-framework/issues/26

### DIFF
--- a/src/Factory/FactoryInterface.php
+++ b/src/Factory/FactoryInterface.php
@@ -23,7 +23,7 @@ interface FactoryInterface
      * Create an object
      *
      * @param  string             $requestedName
-     * @param  null|array         $options
+     * @param  null|array<mixed>  $options
      * @return object
      * @throws ServiceNotFoundException If unable to resolve the service.
      * @throws ServiceNotCreatedException If an exception is raised when creating a service.


src/Factory/FactoryInterface.php
```diff
@@ -23,7 +23,7 @@ interface FactoryInterface
      * Create an object
      *
      * @param  string             $requestedName
-     * @param  null|array         $options
+     * @param  null|array<mixed>  $options
```
Ocramius, line 26: `array` is an alias of `array<int|string, mixed>`? Why is this needed explicitly? :thinking:

Ocramius (changes requested):

Ocramius (approved):
Hmm, I see https://github.com/laminas/laminas-servicemanager/pull/99#discussion_r680998436 is about strictness: from my PoV it's OK to declare it `array<int|string, mixed>`, but it seems weird that this is causing issues for subtype/implementations.

Ocramius:
Thanks @fezfez!

src/Factory/FactoryInterface.php
fezfez: Yeah, that's what i was thinking. Maybe i should report to phpstan instead.

Ocramius:
For others wondering whether this is necessary at all: it isn't, but you can always go back from `array<mixed>` to just `array`, so it doesn't hurt anyway.

boesing:
Actually, `options` should be `array<string,mixed>` as options supposed to be a map.
Thats what I've added in #98 btw.

Ocramius:
@boesing is it really restricted to a map? :thinking:

boesing:
@Ocramius its not but just because it is not declared at all and nobody cares.

Options are mostly passed to plugins where they are either stored by using setters on the same plugin (e.g. validators, filters) or in case of the cache component, passed to the `AdapterOptions`. It just dont make any sense to allow integer keys for an Options Argument...

Ocramius:
@boesing I would expect this to be valid though:

```php
$sevenDwarvesFacility = $pluginManager->create(MiningFacility::class, [$dopey, $grumpy, $bashful, $sleepy, $sneezy]);
```

boesing:
@Ocramius so you expect seven dwarves for 5 variables passed as options? 🤪

I see the point but I would expect passing these variables as an Option "dwarves". 
But whatever, gonna update my PR - has conflicts now anyways.

I would prefer letting the plugin manager to declare what type is expected. Gonna try something with generics, maybe I get something that might work.

Ocramius:
I was just making an example of something that I consider valid: I rarely use options for service creation, but I don't see big value in restricting it in this location, as it's just a mechanism for parameter passthrough, and we'd only introduce a variance BC break.

PR #100, opened 2021-08-03
Remove check for `php:<7`, since support was dropped

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Since support for PHP < 7.3 was dropped in v3.6 of this library, the check for PHP < 7 can be removed.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -23,7 +23,6 @@ use function array_intersect;
 use function array_key_exists;
 use function array_keys;
 use function class_exists;
-use function explode;
 use function get_class;
 use function gettype;
 use function in_array;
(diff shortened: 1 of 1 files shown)


Ocramius (approved):

Ocramius:
@internalsystemerror could you check these two CS failures, please?

```
Error: Type explode is not used in this file.
Error: Type strpos is not used in this file.
```

internalsystemerror:
> @internalsystemerror could you check these two CS failures, please?
> 
> ```
> Error: Type explode is not used in this file.
> Error: Type strpos is not used in this file.
> ```

Thanks for approving the workflow. Changes made as requested.

snapshotpl (approved):

Ocramius:
Thanks again! :ship:

boesing:
I already see this conflicting with #98 :-( 
I've removed this as well.

Ocramius:
@boesing happens, heh :-\

PR #101, opened 2021-08-05
Remove useless @param and @return tag

samsonasik:
|    Q          |   A
|-------------- | ------
| QA            | yes

`@param` and `@return` can be removed if not needed.


samsonasik:
It seems will conflict with https://github.com/laminas/laminas-servicemanager/pull/98 , closing for now.

PR #103, opened 2021-09-13
Provide PHP 8.1 support

weierophinney:
This patch provides PHP 8.1 support, via the following:

- It **DROPS** support for PHP 7.3 versions.
- It adds `~8.1.0` to the list of allowed PHP versions.
- It bumpes phpbench to the 1.1 series.
- It bumps PHPUnit to the 9.5 series.
- It bumps laminas-coding-standard to the 2.2 series.
- It bumps vfsstream to the 1.6.10alpha series.
- It adds shims for the `ClassReflection`, `MethodReflection`, and `ParameterReflection` classes from laminas-code in order to add `#[ReturnTypeWillVary]` attributes on methods extending internals, and thus allow unit tests to run.

Note: this may mean that laminas-servicemanager will not work out-of-the-box when error reporting is set to `E_ALL` if you are using the `ReflectionBasedAbstractFactory` and/or the `bin/generate-factory-for-class` and `bin/generate-deps-for-config-factory` tools, at least until laminas-code has a release that provides PHP 8.1 support.

### DIFF
--- a/test/TestAsset/laminas-code/ClassReflection.php
+++ b/test/TestAsset/laminas-code/ClassReflection.php
@@ -1,0 +1,225 @@
+<?php
+
+namespace Laminas\Code\Reflection;
+
+use ReflectionClass;
+use ReturnTypeWillChange;
+
+use function array_shift;
+use function array_slice;
+use function array_unshift;
+use function file;
+use function file_exists;
+use function implode;
+use function strstr;
+
+class ClassReflection extends ReflectionClass implements ReflectionInterface
+{
+    /** @var DocBlockReflection|null */
+    protected $docBlock;
+
+    /**
+     * Return the classes DocBlock reflection object
+     *
+     * @return DocBlockReflection|false
+     * @throws Exception\ExceptionInterface When missing DocBock or invalid reflection class.
+     */
+    public function getDocBlock()
+    {
+        if (isset($this->docBlock)) {
+            return $this->docBlock;
+        }
+
+        if ('' == $this->getDocComment()) {
+            return false;
+        }
+
+        $this->docBlock = new DocBlockReflection($this);
+
+        return $this->docBlock;
+    }
+
+    /**
+     * Return the start line of the class
+     *
+     * @param  bool $includeDocComment
+     * @return int
+     */
+    #[ReturnTypeWillChange]
+    public function getStartLine($includeDocComment = false)
+    {
+        if ($includeDocComment && $this->getDocComment() != '') {
+            return $this->getDocBlock()->getStartLine();
+        }
+
+        return parent::getStartLine();
+    }
+
+    /**
+     * Return the contents of the class
+     *
+     * @param  bool $includeDocBlock
+     * @return string
+     */
+    public function getContents($includeDocBlock = true)
+    {
+        $fileName = $this->getFileName();
+
+        if (false === $fileName || ! file_exists($fileName)) {
+            return '';
+        }
+
+        $filelines = file($fileName);
+        $startnum  = $this->getStartLine($includeDocBlock);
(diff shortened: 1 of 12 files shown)


test/TestAsset/laminas-code/ClassReflection.php
```diff
@@ -0,0 +1,225 @@
...
+
+use function array_shift;
+use function array_slice;
+use function array_unshift;
+use function file;
+use function file_exists;
+use function implode;
+use function strstr;
+
+class ClassReflection extends ReflectionClass implements ReflectionInterface
```
Ocramius, line 22: Should be documented as temporary shim

test/TestAsset/laminas-code/MethodReflection.php
```diff
@@ -0,0 +1,460 @@
...
+use function rtrim;
+use function strlen;
+use function substr;
+use function token_get_all;
+use function token_name;
+use function var_export;
+
+use const FILE_IGNORE_NEW_LINES;
+
+class MethodReflection extends PhpReflectionMethod implements ReflectionInterface
```
Ocramius, line 31: Should be documented as temporary shim

test/TestAsset/laminas-code/ParameterReflection.php
```diff
@@ -0,0 +1,123 @@
...
+namespace Laminas\Code\Reflection;
+
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionParameter;
+use ReturnTypeWillChange;
+
+use function method_exists;
+
+class ParameterReflection extends ReflectionParameter implements ReflectionInterface
```
Ocramius, line 18: Should be documented as temporary shim

test/autoload.php
```diff
@@ -0,0 +1,11 @@
+<?php // phpcs:disable Generic.Files.LineLength.TooLong
+
+if (PHP_VERSION_ID >= 80100) {
+    require __DIR__ . '/TestAsset/laminas-code/ParameterReflection.php';
+    require __DIR__ . '/TestAsset/laminas-code/MethodReflection.php';
+    require __DIR__ . '/TestAsset/laminas-code/ClassReflection.php';
+
+    class_alias(\Laminas\Code\Reflection\ParameterReflection::class, \Zend\Code\Reflection\ParameterReflection::class);
```
Ocramius: Are these needed at all? In theory, we don't use the `Zend\` aliases at all

Ocramius (changes requested):

test/autoload.php
weierophinney: Yes - when running 8.1 against lowest, there's some oddities that happen because the ocramius proxy-manager version is sufficiently low in order to allow usage with 7.3 in this library.

Ocramius: do we need usage with 7.3? I checked, and that support was for proxymanager from an ancient past...

weierophinney: And, as noted, this is only for testing.

weierophinney: We could potentially bump it to 7.4 as the minimum. I've not done that with any of the other packages we've done 8.1 releases for, though. :slightly_frowning_face:

weierophinney: Bumped PHP minimum to 7.4. :muscle:

boesing:
> It DROPS support for PHP 7.4 versions.

I guess thats a typo?

weierophinney:
> > It DROPS support for PHP 7.4 versions.
> 
> I guess thats a typo?

Ooops! Updated!

Ocramius (approved):
LGTM! @boesing I would like to also merge #98, but IMO best to do an isolated release with only PHP 8.1 support first

PR #104, opened 2021-09-16
[PHP 7.4] Apply Typed Property

samsonasik:
Since we require php 7.4, this PR is to apply typed property.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -54,7 +54,7 @@ use const E_USER_DEPRECATED;
 class ServiceManager implements ServiceLocatorInterface
 {
     /** @var Factory\AbstractFactoryInterface[] */
-    protected $abstractFactories = [];
+    protected array $abstractFactories = [];
 
     /**
      * A list of aliases
@@ -67,16 +67,13 @@ class ServiceManager implements ServiceLocatorInterface
 
     /**
      * Whether or not changes may be made to this instance.
-     *
-     * @var bool
      */
-    protected $allowOverride = false;
+    protected bool $allowOverride = false;
 
-    /** @var ContainerInterface */
-    protected $creationContext;
+    protected ContainerInterface $creationContext;
 
     /** @var string[][]|Factory\DelegatorFactoryInterface[][] */
-    protected $delegators = [];
+    protected array $delegators = [];
 
     /**
      * A list of factories (either as string name or callable)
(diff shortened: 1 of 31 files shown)


src/ServiceManager.php
```diff
@@ -67,16 +68,13 @@ class ServiceManager implements ServiceLocatorInterface
 
     /**
      * Whether or not changes may be made to this instance.
-     *
-     * @var bool
      */
-    protected $allowOverride = false;
+    protected bool $allowOverride = false;
```
samsonasik, line 72: @boesing it seems make error in `pslam --stepherd`

```
Error: Type false for $this->allowOverride is always falsy (see https://psalm.dev/122)
```

Do you have idea to handle it?

boesing: Not really. 

Maybe the reason why this is an error is, that there is no way of changing `allowOverride` via codepath as it only is overridden in plugin managers.
So psalm thinks "this will never change to an other value than `false`.

But thats a weird assumption as it is `protected`. Maybe thats a bug?

samsonasik: probably, as add typed parameter and return type in `setAllowOverride()` and `getAllowOverride()` seems stil got the error:

Screen Shot 2021-09-16 at 21 33 00

samsonasik: any way we can ignore the error?

boesing: Ah, the issue is that `allowOverride` is already verified in the method earlier:

```
if ($this->allowOverride || ! $this->configured) {
    return;
}
```

So thats why you can just remove these checks at all (but keep the initial check).

boesing:
Adding property types is a BC break and should target v4

https://3v4l.org/spl4P

src/ServiceManager.php
samsonasik: Fixed 🎉 https://github.com/laminas/laminas-servicemanager/pull/104/commits/6d2438235fcec2792e069bc9eb236647cebca5ad thank you

samsonasik:
Do you mean it needs to target `dev-4.0` branch? It is is the initial part for property only, next is make param and return typed.

samsonasik:
All green 🎉

boesing:
Yes, it has to target v4. 
Tests are green but would break upstream implementations such as `laminas-filter`, `laminas-form`, `laminas-cache` and all others.

As v4 needs plenty of work, there wont be a v4 any time soon. That was already discussed in many other issues.

https://github.com/laminas/laminas-servicemanager/issues/18#issuecomment-879182790

samsonasik:
`dev-4.0` seems needs rebase first as latest update of `dev-4.0` is `Jan 8, 2020`

Ocramius:
Don't know if the `dev-4.0` branch is in use at all 🤔

Ocramius:
Another suggestion: let's apply typed properties to `private` state (only), and perhaps get this into `3.x`?

samsonasik:
I will apply typed property for private property only in separate PR

samsonasik:
I am closing it, as typed property for private property only already applied in https://github.com/laminas/laminas-servicemanager/pull/106

Ocramius:
@samsonasik please update milestones/labels/etc when closing stuff :)

samsonasik:
@Ocramius I forgot about it, thank you ;)

PR #112, opened 2022-01-23
Add compatibility with renovatebot

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

These changes will allow renovate to work as intended using the new shared configuration from https://github.com/laminas/.github/pull/19.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -4,9 +4,6 @@ on:
   pull_request:
   push:
     branches:
-      - '[0-9]+.[0-9]+.x'
-      - 'refs/pull/*'
-      - 'renovate/*'
 
 jobs:
   matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches:
       - '[0-9]+.[0-9]+.x'
       - 'refs/pull/*'
+      - 'renovate/*'
```
internalsystemerror: @Ocramius You mentioned removing the filter instead. Would I just make this an empty array to do that?

Ocramius: Just removing these 3 lines suffices

internalsystemerror: done.

Ocramius (approved):
💪 thanks @internalsystemerror!

PR #125, opened 2022-04-07
Conflict with `ext-psr`

### LINKED ISSUE #124: Version 3.11 conflicts with ext-psr
When [psr](https://pecl.php.net/package/psr) extension is loaded

`Fatal error: Uncaught ValueError: class_alias(): Argument #1 ($class) must be a user-defined class name, internal class name given in ......src/autoload.php:13
`

so `class_alias` cannot be used there

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | kinda
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | probably

### Description

There is a PHP extension around which provides all (most?) PSR interfaces without the need of installing them via composer. Since we do not support this extension and since that extension does not make sense anyways (more details later), we explicitly adding a `conflict` to the `composer.json` so that there are no surprises for projects which do have the `psr` extension installed.

closes #124

Now to the problem with the extension:

There are multiple PSR standards around. Many of them do have multiple major versions around. The extension has to follow semver to avoid breaks in upstream projects. So as of now (v1.2.0), it only provides the interfaces for the v1 versions of the PSR standards.
There is no way to provide multiple major versions of the same PSR interface from within one version of the extension.
Therefore, at some point, an upstream project is either stuck with the v1 versions of the PSR standards or the extension needs to create a hilarious amount of major versions to reflect all possible combinations of all major versions (cartesian product).

Lets just take the PSR-11 + PSR-6 example, we already would have 6 major versions around because PSR-11 has 2 major versions and PSR-6 has 3 major versions:

- psr/cache 1.x + psr/container 1.x
- psr/cache 1.x + psr/container 2.x
- psr/cache 2.x + psr/container 1.x
- psr/cache 2.x + psr/container 2.x
- psr/cache 3.x + psr/container 1.x
- psr/cache 3.x + psr/container 2.x

And these are only 2 packages. I lost track on the current progress, but there will be more major versions in the future and thus the amount of major versions for the extension will grow (and with that, the confusion for upstream projects, maintainers, etc.).

We highly recommend to use composer to install the PSR standards per package in the version the upstream project can work with.
**Do not use the `psr` extension**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,6 +53,7 @@
         "psr/container-implementation": "^1.0"
     },
     "conflict": {
+        "ext-psr": "*",
         "laminas/laminas-code": "<3.3.1",
         "zendframework/zend-code": "<3.3.1",
         "zendframework/zend-servicemanager": "*"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e238a4c88ecd1972f68bc95830a9f920",
+    "content-hash": "aa071135de48212d7c9eba1a27bcbe1c",
     "packages": [
         {
             "name": "laminas/laminas-stdlib",
@@ -5431,5 +5431,5 @@
     "platform-overrides": {
         "php": "7.4.99"
     },
-    "plugin-api-version": "2.2.0"
+    "plugin-api-version": "2.3.0"
 }


Ocramius:
Note to readers: `ext-psr` makes sense when PSR classes are needed in extension code, but as it currently stands, that package is a hot mess, and we can't support it until we just declare `psr/container` as a dependency here (and it will take a while).

composer.json
```diff
@@ -53,6 +53,7 @@
         "psr/container-implementation": "^1.0"
     },
     "conflict": {
+        "ext-psr": "*",
```
Ocramius, line 56: `composer.lock` needs updating, no?

Ocramius (changes requested):

composer.json
boesing: good catch. conflcit is not added to the `composer.lock`, only the `composer-plugin-version`.
Not sure if its worth to change the lock-file just to bump the plugin API to 2.3.0 tho.
WDYT?

boesing: Updated.

Ocramius (approved):
Thanks @boesing!

PR #129, opened 2022-04-26
Fix issue with opcache preload and class_alias

### LINKED ISSUE #128: Cannot declare interface Interop\Container\Exception\NotFoundException, because the name is already in use
### BC Break Report

|    Q        |   A
|------------ | ------
| Version     | 3.11
| PHP-FPM with opcache.preload enabled | 8.1.4
| Symfony  (with Runtime) | 5.4.7

#### Summary
After upgrading to version 3.11 Im getting all these E_WARNING messages in my logs and monitoring tools.  Im using Symfony 5.4 and FPM with opcache.preload enabled. When I disable preload the warning goes away. 

```
E_WARNING: Cannot declare interface Interop\Container\Containerinterface, because the name is already in use
in class_alias called at /app/vendor/laminas/laminas-servicemanager/src/autoload.php (13)
in require called at /app/vendor/composer/autoload_real.php (55)
in composerRequire6635dcbc130fab9e00890c611d69ace8 called at /app/vendor/composer/autoload_real.php (38)
in ComposerAutoloaderInit6635dcbc130fab9e00890c611d69ace8::getLoader called at /app/vendor/autoload.php (12)
in require_once called at /app/vendor/autoload_runtime.php (5)
in require_once called at /app/public/index.php (9)
```
And also for line 14/15 I get the warning about ContainerException/NotFoundException.

By upgrading to version 3.11 the Interop package got removed. I have no other references to Interop in my generated preload files or any other place. I think its related to opcache.preload, but im not sure.

In a previous version of the src/autoload.php file, there was an `class_exists` check before the alias:
https://github.com/laminas/laminas-servicemanager/commit/0732d65c14962d4aa25c78788cae5bb0a4ac537f#diff-984b8e3111197dd008894b4b1575ffcbda5c801cb8857434cc2898b5dca67873R13

But when I use that version, the warning is still present. 

Downgrading to 3.10.* resolves the issue.

starred-gijs:
Fixes #128 

|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Not sure how to test this, except the manual tests I ran.

### DIFF
--- a/src/autoload.php
+++ b/src/autoload.php
@@ -10,6 +10,13 @@ use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\ContainerInterface;
 use Psr\Container\NotFoundExceptionInterface;
 
-class_alias(ContainerInterface::class, InteropContainerInterface::class);
-class_alias(ContainerExceptionInterface::class, InteropContainerException::class);
-class_alias(NotFoundExceptionInterface::class, InteropNotFoundException::class);
+if (! interface_exists(InteropContainerInterface::class, false)) {
+    class_alias(ContainerInterface::class, InteropContainerInterface::class);
+}
+if (! interface_exists(InteropContainerException::class, false)) {
+    class_alias(ContainerExceptionInterface::class, InteropContainerException::class);
+}
+
+if (! interface_exists(InteropNotFoundException::class, false)) {
+    class_alias(NotFoundExceptionInterface::class, InteropNotFoundException::class);
+}


boesing:
The problem is, that the whole idea of replacing `container-interop` is to replace it from within THIS repository.
If we now allow other interfaces to replace `container-interop`, the whole idea of why we made this change won't work anymore. Because your application will break for sure in case that the `Interop\ContainerInterop\ContainerInterface` is not **exactly** the same interface as `Psr\Container\ContainerInterface`.

---

So do I understand correct, that calling `class_alias` in an opcached application breaks when this was done after the initial loading of that script?

When we do add `interface_exists` here, we should at least verify that the interface which does exist is pointing the expected PSR-11 interface.

Something like this (which does obv. not work, but I don't know how to verify that):
```php
interface Foo
{}
if (!interface_exists(Bar::class)) {
     class_alias(Foo::class, 'Bar');
}

assert(Bar::class === Foo::class);
```

Because silently allow projects to have an existing implementation of the interop interface might break projects as we changed the method type-hints to require PSR-11 implementations.

__(Is it clear what I want to point out? I am not sure if that was clear...)__

Ocramius:
That's kinda why I wanted to test this: I think this might be a limitation of preloading, which only stores the declarations.

PR #133, opened 2022-06-03
Fixed namespace generation in `FactoryCreator` utility

adambalint-srg:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

In FactoryCreator's createFactory method the namespace of the class has been built with `str_replace`. This method worked for this kind of classes: `Foo\Bar\Service`, but the generated namespace was wrong in case of this kind of classes: `Foo\Service\Bar\Service`, because namespace contains the class name too.

| FQCN | Generated namespace
|-|-
| Foo\Bar\Service | Foo\Bar
| Foo\Service\Bar\Service | Foo\Bar

With `preg_replace` it is possible to cut off only the last part of the FQCN.

### DIFF
--- a/src/Tool/FactoryCreator.php
+++ b/src/Tool/FactoryCreator.php
@@ -16,10 +16,10 @@ use function array_merge;
 use function array_shift;
 use function count;
 use function implode;
+use function preg_replace;
 use function sort;
 use function sprintf;
 use function str_repeat;
-use function str_replace;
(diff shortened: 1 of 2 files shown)


Ocramius:
Thanks @adambalint-srg!

PR #134, opened 2022-06-03
Update the Reflection Factory documentation

settermjd:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

After working through the documentation to learn more about it, I felt that the documentation could be refined to make it that much easier to read, as well as to extend the usage examples to make them more encompassing of different use cases.

PR #136, opened 2022-06-08
Update reflection based abstract factory documentation

settermjd:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

After working through the documentation to learn more about the Reflection-based Abstract Factor, I felt that the documentation could be refined to make it that much easier to read and maintain, as well as to extend the usage examples to make them more encompassing of different use cases. This PR replaces https://github.com/laminas/laminas-servicemanager/pull/134.

### DIFF
--- a/docs/book/reflection-abstract-factory.md
+++ b/docs/book/reflection-abstract-factory.md
@@ -18,7 +18,7 @@ The factory operates with the following constraints/features:
 
 WARNING: `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.
 
-Once your dependencies have stabilized, we recommend writing a dedicated factory, as reflection can introduce performance overhead. 
+Once your dependencies have stabilized and when performance is a requirement, we recommend writing a dedicated factory, as reflection can introduce performance overhead. 
 For example, you could use the [generate-factory-for-class console tool](console-tools.md#generate-factory-for-class) to do so.
 
 ## Usage Example
--- a/docs/book/reflection-abstract-factory.md
+++ b/docs/book/reflection-abstract-factory.md
@@ -2,35 +2,24 @@
 
 - Since 3.2.0.
 
-Writing a factory class for each and every service that has dependencies
-can be tedious, particularly in early development as you are still sorting
-out dependencies.
+Writing a factory class for each and every service that has dependencies can be tedious, particularly in early development as you are still sorting out dependencies.
 
-To alleviate this issue during development, laminas-servicemanager ships with the [ReflectionBasedAbstractFactory](https://github.com/laminas/laminas-servicemanager/blob/3.12.x/src/AbstractFactory/ReflectionBasedAbstractFactory.php),
-which provides a [reflection-based approach](https://www.php.net/manual/en/intro.reflection.php) to instantiation, 
-resolving constructor dependencies to the relevant services. 
+To alleviate this issue during development, laminas-servicemanager ships with the `ReflectionBasedAbstractFactory`, which provides a [reflection-based approach](https://www.php.net/manual/en/intro.reflection.php) to instantiation, resolving constructor dependencies to the relevant services. 
 The factory may be used as either an abstract factory or mapped to specific service names as a factory.
 
 TIP: Mapping services to the factory is more explicit and performant.
 
 The factory operates with the following constraints/features:
 
-- A parameter named `$config` type-hinted as an array will receive the
-  application "config" service (i.e., the merged configuration).
-- Parameters type-hinted against array, but not named `$config`, will
-  be injected with an empty array.
-- Scalar parameters will result in the factory raising an exception,
-  unless a default value is present; if it is, that value will be used.
-- If a service cannot be found for a given typehint, the factory will
-  raise an exception detailing this.
+- A parameter named `$config` type-hinted as an array will receive the application "config" service (i.e., the merged configuration).
+- Parameters type-hinted against array, but not named `$config`, will be injected with an empty array.
+- Scalar parameters will result in the factory raising an exception, unless a default value is present; if it is, that value will be used.
+- If a service cannot be found for a given typehint, the factory will raise an exception detailing this.
 
-WARNING: `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which 
-argument(s) they might replace.
+WARNING: `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.
 
-Once your dependencies have stabilized, we recommend writing a dedicated
-factory, as reflection can introduce performance overhead. For example, you could use the
-[generate-factory-for-class console tool](console-tools.md#generate-factory-for-class)
-to do so.
+Once your dependencies have stabilized, we recommend writing a dedicated factory, as reflection can introduce performance overhead. 
+For example, you could use the [generate-factory-for-class console tool](console-tools.md#generate-factory-for-class) to do so.
 
 ## Usage Example
 
@@ -85,8 +74,8 @@ public function getDependencies(): array
 
 ## Handling well-known services
 
-Some services provided by Laminas components do not have
-entries based on their class name (for historical reasons). As examples:
+Some services provided by Laminas components do not have entries based on their class name (for historical reasons). 
+As examples:
 
 - `Laminas\Console\Adapter\AdapterInterface` maps to the service name `ConsoleAdapter`,
 - `Laminas\Filter\FilterPluginManager` maps to the service name `FilterManager`,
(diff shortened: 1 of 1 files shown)


Ocramius:
Note: IMO for `3.13.x`

settermjd:
> Note: IMO for `3.13.x`

I'm pretty sure @froschdesign recommended 3.11. @froschdesign?

Ocramius:
Unless it's a bugfix, it should be landing in new minors in general :D

froschdesign:
> Note: IMO for `3.13.x`

As already mentioned: with the current state of documentation, in terms of quality and quantity, any change can be considered as a **patch**. No matter if something is updated or added.
I have no problem with a patch for the documentation being released as a minor version with other pull requests. But only a change to the documentation as a minor makes the release rather sad. Because nothing is changed on the user side after a Composer update.

Ocramius:
I'd do a patch release even for docs, which is just more noise tho? :thinking: 

In this case, I'd do 2 releases + 2 merge-ups

froschdesign:
> I'd do a patch release even for docs, which is just more noise tho?

It would be much better if changes to documentation did not require releases at all. :stuck_out_tongue_winking_eye:

Ocramius:
I don't mind, honestly - we don't have enough noise from it for it to be a problem.

settermjd:
@Ocramius, if you don't mind either way, I'd prefer to leave it as is. On a different note, what can I do about the DCO check failing, as it relates to laminas bot?
Bildschirmfoto vom 2022-06-17 12-51-36
?

Just merge anyway?

Ocramius:
Yeah, DCO is useless corporateism, so it can be ignored on anything that isn't "world-changing tech", TBH.

Ocramius:
@froschdesign your call on where/how to merge/release

docs/book/reflection-abstract-factory.md
```diff
@@ -2,14 +2,33 @@
...
+Writing a factory class for each and every service that has dependencies can be tedious, particularly in early development as you are still sorting out dependencies.
 
-laminas-servicemanager ships with `Laminas\ServiceManager\AbstractFactory\ReflectionBasedAbstractFactory`,
-which provides a reflection-based approach to instantiation, resolving
-constructor dependencies to the relevant services. The factory may be used as
-either an abstract factory, or mapped to specific service names as a factory:
+To alleviate this issue during development, laminas-servicemanager ships with the `ReflectionBasedAbstractFactory`, which provides a [reflection-based approach](https://www.php.net/manual/en/intro.reflection.php) to instantiation, resolving constructor dependencies to the relevant services. 
+The factory may be used as either an abstract factory or mapped to specific service names as a factory.
+
+TIP: Mapping services to the factory is more explicit and performant.
```
froschdesign, line 10: I would move the tip to the end of this section.

```suggestion
```

docs/book/reflection-abstract-factory.md
```diff
@@ -2,14 +2,33 @@
...
+
+The factory operates with the following constraints/features:
+
+- A parameter named `$config` type-hinted as an array will receive the application "config" service (i.e., the merged configuration).
+- Parameters type-hinted against array, but not named `$config`, will be injected with an empty array.
+- Scalar parameters will result in the factory raising an exception, unless a default value is present; if it is, that value will be used.
+- If a service cannot be found for a given typehint, the factory will raise an exception detailing this.
+
+WARNING: `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.
```
froschdesign, line 19: The list is introduced with "features" and "constraints" therefore this warning can be a normal list entry:

```suggestion
- If a service cannot be found for a given typehint, the factory will raise an exception detailing this.
- `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.
```

docs/book/reflection-abstract-factory.md
```diff
@@ -2,14 +2,33 @@
...
-laminas-servicemanager ships with `Laminas\ServiceManager\AbstractFactory\ReflectionBasedAbstractFactory`,
-which provides a reflection-based approach to instantiation, resolving
-constructor dependencies to the relevant services. The factory may be used as
-either an abstract factory, or mapped to specific service names as a factory:
+To alleviate this issue during development, laminas-servicemanager ships with the `ReflectionBasedAbstractFactory`, which provides a [reflection-based approach](https://www.php.net/manual/en/intro.reflection.php) to instantiation, resolving constructor dependencies to the relevant services. 
+The factory may be used as either an abstract factory or mapped to specific service names as a factory.
+
+TIP: Mapping services to the factory is more explicit and performant.
+
+The factory operates with the following constraints/features:
```
froschdesign, line 12: The positive first:

```suggestion
The factory operates with the following features/constraints:
```

docs/book/reflection-abstract-factory.md
```diff
@@ -2,14 +2,33 @@
...
+WARNING: `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.
+
+Once your dependencies have stabilized and when performance is a requirement, we recommend writing a dedicated factory, as reflection can introduce performance overhead. 
+For example, you could use the [generate-factory-for-class console tool](console-tools.md#generate-factory-for-class) to do so.
+
+## Usage Example
+
+### Using Laminas MVC
+
+When using [Laminas MVC](https://docs.laminas.dev/mvc/):
```
froschdesign, line 28: The package name for laminas-mvc is used most and is also the preferred way so far.
(If something refers to Mezzio, then the spelling with a capital initial is used: "Mezzio". The package names also follow the names used in Composer. Example: "mezzio-problem-details".)

```suggestion
### Using laminas-mvc

When using [laminas-mvc](https://docs.laminas.dev/mvc/):
```

docs/book/reflection-abstract-factory.md
```diff
@@ -2,14 +2,33 @@
...
+
+- Add the factory in the service manager's list of abstract factories.
+- Use the `ReflectionBasedAbstractFactory` to instantiate the class.
```
froschdesign, line 31: This part and also the code example are wrong because you do not need both options to use `ReflectionBasedAbstractFactory`.

- If the `ReflectionBasedAbstractFactory` is used as an abstract factory, then it will do the job for all services to which no explicit factory is mapped.
- If the `ReflectionBasedAbstractFactory` is used as factory for a single service, then it will work only for this one service.

This means that we need two _independent_ examples for the documentation, each with its own explanation!

docs/book/reflection-abstract-factory.md
```diff
@@ -28,31 +47,35 @@ return [
...
+        'abstract_factories' => [
+            ReflectionBasedAbstractFactory::class,
+        ],
+        'factories' => [
+            // Other factories...
+            Handler\HomePageHandler::class => ReflectionBasedAbstractFactory::class,
+        ],
+    ],
+}
+```
```
froschdesign, line 73: The same as above: this is wrong.

docs/book/reflection-abstract-factory.md
```diff
@@ -2,14 +2,33 @@
...
+
+- A parameter named `$config` type-hinted as an array will receive the application "config" service (i.e., the merged configuration).
+- Parameters type-hinted against array, but not named `$config`, will be injected with an empty array.
+- Scalar parameters will result in the factory raising an exception, unless a default value is present; if it is, that value will be used.
+- If a service cannot be found for a given typehint, the factory will raise an exception detailing this.
+
+WARNING: `$options` passed to the factory are ignored in all cases, as we cannot make assumptions about which argument(s) they might replace.
+
+Once your dependencies have stabilized and when performance is a requirement, we recommend writing a dedicated factory, as reflection can introduce performance overhead. 
+For example, you could use the [generate-factory-for-class console tool](console-tools.md#generate-factory-for-class) to do so.
```
froschdesign, line 22: I would suggest adding the tip here:

```suggestion
TIP: Once your dependencies have stabilized and when performance is a requirement, we recommend writing a dedicated factory, as reflection can introduce performance overhead. 
For example, you could use the [generate-factory-for-class console tool](console-tools.md#generate-factory-for-class) to do so.
```

froschdesign (changes requested):
_Sorry for the delay!_

Good improvements, but the current usage example needs an update and clarification.

boesing:
@settermjd @froschdesign I've also modified the documentation in #180 

Should we try to finish this as well and then merge-up to 4.0.x?

froschdesign:
@boesing 
> Should we try to finish this as well and then merge-up to 4.0.x?

A good idea, because the entire documentation needs to be reworked for version 4.0.

PR #137, opened 2022-06-13
Improve type inference for plugin manager implementations

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Improves type inference particularly for plugin managers in dependent libs

### DIFF
--- a/src/ServiceLocatorInterface.php
+++ b/src/ServiceLocatorInterface.php
@@ -15,9 +15,11 @@ interface ServiceLocatorInterface extends ContainerInterface
     /**
      * Build a service by its name, using optional options (such services are NEVER cached).
      *
-     * @param  string $name
+     * @template T of object
+     * @param  string|class-string<T> $name
      * @param  null|array<mixed>  $options
      * @return mixed
+     * @psalm-return ($name is class-string ? T : mixed)
      * @throws Exception\ServiceNotFoundException If no factory/abstract
      *     factory could be found to create the instance.
      * @throws Exception\ServiceNotCreatedException If factory/delegator fails
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -259,6 +259,9 @@ class ServiceManager implements ServiceLocatorInterface
 
     /**
      * {@inheritDoc}
+     *
+     * @param string|class-string $name
+     * @return bool
      */
     public function has($name)
     {
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.22.0@fc2c6ab4d5fa5d644d8617089f012f3bb84b8703">
+<files psalm-version="4.23.0@f1fe6ff483bf325c803df9f510d09a03fd796f88">
   <file src="src/AbstractFactory/ConfigAbstractFactory.php">
     <InvalidStringClass occurrences="1">
       <code>new $requestedName(...$arguments)</code>
@@ -405,7 +405,7 @@
       <code>$names[$name]</code>
       <code>$object[$shared ? $method : 'build']</code>
     </MixedArrayOffset>
-    <MixedAssignment occurrences="26">
+    <MixedAssignment occurrences="24">
       <code>$callSequence</code>
       <code>$first</code>
       <code>$idx1</code>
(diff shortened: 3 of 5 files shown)


boesing:
Just a side-note here:

I was about adding this in #98 
Back then, @Ocramius told me that this would be a soft BC break due to the way how generics work. So when generic declaration is missing, this will become very messy in projects after upgrading to the version introducing generics.

Therefore, I deferred that to servicemanager v4

src/ServiceLocatorInterface.php
```diff
@@ -15,9 +15,11 @@ interface ServiceLocatorInterface extends ContainerInterface
     /**
      * Build a service by its name, using optional options (such services are NEVER cached).
      *
-     * @param  string $name
+     * @template T of object
+     * @param  string|class-string<T> $name
      * @param  null|array<mixed>  $options
      * @return mixed
+     * @psalm-return ($name is class-string ? T : mixed)
```
Ocramius, line 22: In theory, yes.
In practice, this is not necessarily true.

I think adding type inference is good though, even if it **may** be a lie :thinking: 

It's an accepted convention, so I'd say :+1:, and whoever breaks the convention is at fault here instead.

src/ServiceManager.php
```diff
@@ -259,6 +259,9 @@ public function build($name, ?array $options = null)
 
     /**
      * {@inheritDoc}
+     *
+     * @param string|class-string $name
```
Ocramius, line 263: `string` contains `class-string` :thinking:

Ocramius (approved):
LGTM - should we :ship: ?

Ocramius:
> Back then, @Ocramius told me that this would be a soft BC break due to the way how generics work. 

I think this only applies to our `latest` CI builds, where `psalm` would complain about a non-provided template parameter?

I'd say this is good improvement, and doesn't affect the API per se, just documentation. Downstream consumers that use psalm are used to upgrades "shifting" the types a bit, and will adjust.

boesing:
> I'd say this is good improvement, and doesn't affect the API per se, just documentation. Downstream consumers that use psalm are used to upgrades "shifting" the types a bit, and will adjust.

I'm fine with this. 

Anyways: we need to adjust our own components as well, which are quite a lot.
Can we probably create a project for this and list all components which need work then?

Thinking about:
- input filters
- validators
- cache adapters
- cache plugins (or did I killed that? not 100% sure...)
- form elements
- etc.

_(btw: with `we` I mean "not me" 😄)_

Ocramius:
> Can we probably create a project for this and list all components which need work then?

I wonder if the installation of renovate (which didn't happen yet, I know, sorry :( ) would already handle this?

boesing:
> I wonder if the installation of renovate would already handle this?

Since I don't think that renovate will create a list of components where the update to the latest `servicemanager` version fails, I'd say no. And I doubt that all the affected projects do have `renovate` setup anyways.

Ocramius:
Well, `psalm` would fail there, and renovate would open a PR :D

froschdesign:
> Can we probably create a project for this and list all components which need work then?

I can create the project to track the progress and to give an overview.

Ocramius:
@froschdesign please do: I just tried creating one, and it messed up everything, possibly because I have 2 versions of GitHub Projects on my system (the pre-release one too)

src/ServiceManager.php
gsteel: As documentation, this reads to me like - "I might get a regular string literal or a class-string" - is that not an improvement on plain old string?

Ocramius: No strong feelings, but other tools do indeed report this as a redundant type union

src/ServiceLocatorInterface.php
Ocramius: Yeh, let's go with it 👍

Ocramius:
Meanwhile, :ship: here.

src/ServiceLocatorInterface.php
Slamdunk: If we got it for `build`, that makes sense for `get` too, doesn't it?
`psr/container` would not accept it, but we can do it :wink:

froschdesign:
@gsteel @Ocramius @boesing 
With some delay, here is a first start: https://github.com/orgs/laminas/projects/21

PR #141, opened 2022-06-28
Skip `class_alias()` declarations when interfaces were already declared, allowing for usage with preloading

### LINKED ISSUE #128: Cannot declare interface Interop\Container\Exception\NotFoundException, because the name is already in use

LucasHantz:
Signed-off-by: Lucas Hantz <lucas.hantz@gmail.com>

Fixes #128 
Closes #143

### DIFF
--- a/src/autoload.php
+++ b/src/autoload.php
@@ -1,5 +1,6 @@
 <?php // phpcs:disable WebimpressCodingStandard.PHP.CorrectClassNameCase.Invalid
 
+
 declare(strict_types=1);
 
 use Interop\Container\Containerinterface as InteropContainerInterface;
@@ -9,12 +10,12 @@ use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\ContainerInterface;
 use Psr\Container\NotFoundExceptionInterface;
 
-if (!interface_exists(InteropContainerInterface::class, false)) {
+if (! interface_exists(InteropContainerInterface::class, false)) {
     class_alias(ContainerInterface::class, InteropContainerInterface::class);
 }
-if (!interface_exists(InteropContainerException::class, false)) {
+if (! interface_exists(InteropContainerException::class, false)) {
     class_alias(ContainerExceptionInterface::class, InteropContainerException::class);
 }
-if (!interface_exists(InteropNotFoundException::class, false)) {
+if (! interface_exists(InteropNotFoundException::class, false)) {
     class_alias(NotFoundExceptionInterface::class, InteropNotFoundException::class);
 }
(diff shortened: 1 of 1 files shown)


Ocramius (changes requested):
@LucasHantz please check inline CI reported errors (you can run `vendor/bin/phpcbf` on that file, if that makes it simpler.

@boesing what's your opinion on this? Should we check if something was autoloaded? This will fix preloading, but it could indeed backfire if somebody else declared those interfaces, and they aren't what we expect (full crash, in that case).

LucasHantz:
Updated with the proper code style

boesing:
@Ocramius TBH: if somebody already used `class_alias` for whatever reason, that might not be our problem.

We either have those users who complain about doing `class_alias` here without a condition or the users who complain about having crashes in their application because whatever other library had used `class_alias` to link to non-PSR interfaces (which might be a super edge-case hopefully).

So lets take that pill and see how that works 🤷🏼‍♂️

boesing (approved):

Ocramius:
Decided to treat this as feature/compatibility with preloading, hence moving this to `3.13.0` :+1: 

Thanks @LucasHantz

Issue #146, opened 2022-07-20
psr/container v2 breaks other components


remicollet:
Since v2 is allowed, example for laminas-inputfilter 

```
$ composer update
...
  - Locking laminas/laminas-servicemanager (3.15.0)
  - Locking psr/container (2.0.2)
...

```

And then

```
$ vendor/bin/phpunit 
PHPUnit 9.5.21 #StandWithUkraine

.............................................................   61 / 1092 (  5%)
.............................................................  122 / 1092 ( 11%)
.............................................................  183 / 1092 ( 16%)
.............................................................  244 / 1092 ( 22%)
.............................................................  305 / 1092 ( 27%)
.............................................................  366 / 1092 ( 33%)
.............................................................  427 / 1092 ( 39%)
.............................................................  488 / 1092 ( 44%)
S...S.......SSSS.............................................  549 / 1092 ( 50%)
.......................................S...S.....SSSS........  610 / 1092 ( 55%)
.............................................................  671 / 1092 ( 61%)
........S.................ESS................................  732 / 1092 ( 67%)
.............................................................  793 / 1092 ( 72%)
.............................................................  854 / 1092 ( 78%)
.............................................................  915 / 1092 ( 83%)
.............................................................  976 / 1092 ( 89%)
............................................................. 1037 / 1092 ( 94%)
.......................................................       1092 / 1092 (100%)

Time: 00:00.099, Memory: 18,00 MB

There was 1 error:

1) LaminasTest\InputFilter\InputFilterPluginManagerFactoryTest::testFactoryReturnsPluginManager
TypeError: Double\ContainerInterface\P8::has(): Return value must be of type bool, null returned

/work/GIT/laminas/laminas-inputfilter/src/InputFilterPluginManagerFactory.php:43
/work/GIT/laminas/laminas-inputfilter/test/InputFilterPluginManagerFactoryTest.php:24

ERRORS!
Tests: 1092, Assertions: 4359, Errors: 1, Skipped: 15.

```

Other components may be affected (laminas-hydrator, laminas-servicemanager-di)
despite these have no direct dependency on psr/container

Ocramius:
Indeed, seen in https://github.com/laminas/laminas-feed/pull/59

It's generally super-annoying, so we're thinking of rolling back `psr/container:^2` support :-\

Ocramius:
Meanwhile, configuring renovate on `laminas-inputfilter`, so this bug pops up in a PR tomorrow morning.

[...]

PR #127, opened 2022-04-07
Support for `psr/container` v2

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes

### Description

Adding support for `psr/container` v2.

### TODO

- [x] bump `psr/container-implementation` in `composer.json`

PR #147, opened 2022-07-20
Revert "Support for `psr/container` v2"

Ocramius:
Reverts laminas/laminas-servicemanager#127

As found in multiple downstream containers, the lack of a constraint on `psr/container:^1` led to upgrades that are completely incompatible with `psr/container:^2` signatures.

The problem is that child classes have methods without a `: bool` return type declaration for `ContainerInterface#has(string $name): bool`, which means that a fatal error is raised, due to LSP compatibility checks performed by PHP.

Practically, upgrading to `psr/container:^2` may really happen in a major version, possibly as an isolated dependency upgrade.

While @boesing has worked hard in trying to implement a BC-compliant solution, the SAT solver of composer will find a way to upgrade to incompatible dependencies anyway, and inheritance is such a massive headache that we may really just want to get rid of it overall, rather than

Ref: https://github.com/laminas/laminas-hydrator/issues/92
Ref: https://github.com/laminas/laminas-di/issues/78
Ref: https://github.com/laminas/laminas-servicemanager/issues/146


boesing (approved):
Simple revert. I don't see a better solution tho.
We should think about closing our implementations while providing proper interfaces to allow composition.

IMHO, v4 should have `final ServiceManager` and the plugin manager should not implement PSR-11 anymore.
If the plugin manager is as configurable as possible, we might not even need an abstract plugin manager.

Ocramius:
Can't wait myself for `final` on the ServiceManager :D

Seldaek:
Due to 3.15.0 existing with psr/container 2 support it means I am now stuck on that version as I have other things requiring psr/container ^2.

I don't think it matters much given the fairly small usage of laminas components I have in the project, but just wanted to let you know that you may be protecting some users from bad upgrades (tho really they can easily pin psr/container:^1 themselves if their code requires it to function) but it is also causing issues for other users that are ready to update.

PR #160, opened 2022-10-26
Explicitly declare that delegator factories support `array<mixed>` for construct-time options, instead of just declaring `array`

fezfez:
same has #99 but for DelegatorFactoryInterface


Ocramius (approved):
LGTM, thanks @fezfez!

PR #164, opened 2022-12-03
WIP: Make `AbstractPluginManager#get()` retrieve `InstanceType` for non-`class-string` input

Ocramius:
This change makes it so that `AbstractPluginManager#get(string): InstanceType` is always produced: since the plugin manager validates each retrieved instance, the type can safely be assumed to **NOT** be `mixed`.

This should improve overall type inference across our components.

Also, this should invalidate calls such as `AbstractPluginManager#get(NotAPlugin::class)`, since the `NotAPlugin&InstanceType` will result in an impossible type anyway.

Note: this entire feature depends on https://github.com/vimeo/psalm/issues/8815. Without
      implementing https://github.com/vimeo/psalm/issues/8815 first, we cannot proceed, as
      downstream consumers will otherwise be able to request any `class-string<T>` to the
      plugin manager system!

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -143,10 +143,11 @@ abstract class AbstractPluginManager extends ServiceManager implements PluginMan
     }
 
     /**
-     * @param class-string<InstanceType>|string $name Service name of plugin to retrieve.
+     * @template RequestedType of object
+     * @param class-string<RequestedType>|string $name Service name of plugin to retrieve.
      * @param null|array<mixed> $options Options to use when creating the instance.
      * @return mixed
-     * @psalm-return ($name is class-string<InstanceType> ? InstanceType : mixed)
+     * @psalm-return ($name is class-string<RequestedType> ? (RequestedType&InstanceType) : InstanceType)
      * @throws Exception\ServiceNotFoundException If the manager does not have
      *     a service definition for the instance, and the service is not
      *     auto-invokable.
--- a/test/StaticAnalysis/AbstractPluginManager.php
+++ b/test/StaticAnalysis/AbstractPluginManager.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\ServiceManager\StaticAnalysis;
+
+use Laminas\ServiceManager\AbstractPluginManager as PluginManager;
+use LaminasTest\ServiceManager\TestAsset\Foo;
+use stdClass;
+
+final class AbstractPluginManager
+{
+    /** @param PluginManager<Foo|stdClass> $pluginManager */
+    public function pluginManagerRetrievesDefaultType(PluginManager $pluginManager): Foo|stdClass
+    {
+        return $pluginManager->get('unknown-type-will-result-in-plugin-manager-default-type');
+    }
+
+    /** @param PluginManager<Foo|stdClass> $pluginManager */
+    public function pluginManagerRetrievesRequestedFooType(PluginManager $pluginManager): Foo
+    {
+        return $pluginManager->get(Foo::class);
+    }
+
+    /** @param PluginManager<Foo|stdClass> $pluginManager */
+    public function pluginManagerRetrievesRequestedStdClassType(PluginManager $pluginManager): stdClass
+    {
+        return $pluginManager->get(stdClass::class);
+    }
+
+    /**
+     * @see https://github.com/vimeo/psalm/issues/8815
+     *
+     * @param PluginManager<Foo|stdClass> $pluginManager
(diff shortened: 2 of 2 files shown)


boesing:
AFAIR, the `instanceOf` property is allowed to be `null` and in case it is `null` or an empty string, no validation is being done.
Do we want to require an `instanceOf` as of v4?

https://github.com/laminas/laminas-servicemanager/blob/f22969b8b0ea2d695aa832c8895be29622b9bd94/src/AbstractPluginManager.php#L191-L193

weierophinney:
This looks perfectly sane to me, @Ocramius. Is the help you need in the form of pulling this branch and testing implementations against it?

Ocramius:
@weierophinney just decision making on this

weierophinney:
@Ocramius So, I guess one of the questions is what might be considered a valid "instanceof"... because we have at least one outlier: laminas-filter. Plugins there can either be an instance of `FilterInterface` OR any callable. If we were to go this route, would it be possible for `$instanceOf` to be a union type — e.g. `callable|FilterInterface`? Or would the individual implementations be able to override this for these particular scenarios?

Ocramius:
Good point: perhaps validation should be delegated to a callback function like `callable(mixed): TPlugin`?

weierophinney:
> perhaps validation should be delegated to a callback function

I mean, that's essentially what the `validate()` function does — it just doesn't return anything (it throws an exception if the instance is invalid for the plugin manager). Maybe we could modify it so it returns the valid instance, which would allow us to put the typing information in the return type of that function?

Ocramius:
> Maybe we could modify it so it returns the valid instance, which would allow us to put the typing information in the return type of that function?

Possible, yes.
My main idea was to prevent the usage of subtyping to create PluginManagers in first place: if our PluginManager accepts a validation callback, we don't need inheritance anymore.

```php
final class PluginManager
{
    public function __construct(ContainerInterface $actualContainer, callable $validate) { ... }
}
```

weierophinney:
> if our PluginManager accepts a validation callback, we don't need inheritance anymore.

ooooh! I like that!

boesing:
> My main idea was to prevent the usage of subtyping to create PluginManagers in first place: if our PluginManager accepts a validation callback, we don't need inheritance anymore.

I'd love that too, thats mainly why I created https://github.com/laminas/laminas-servicemanager/issues/174, but getting rid of the abstract plugin manager might help as well.
The question would be, how to pass the information of the validate callback back to userland as use-case is usually something like:

```php
$filters = $container->get(FilterPluginManager::class);
$filter = $filters->get(MyFilter::class);
```

Does the `FilterPluginManager` then provides native typehints?

Ocramius:
Is `$container->get(FilterPluginManager::class);` a common use-case?

I would probably do `$container->get('PluginManager<MyType>')` or such, although ugly...

boesing:
> Is `$container->get(FilterPluginManager::class);` a common use-case?

Somewhere, you have to fetch the plugin manager to work with. The pluginmanager factory loads the config and adds all filters the project registered via servicemanager specific config (i.e. `factories`, `delegators`, and stuff).

When it comes to chains, you also have to fetch the plugin manager and set it to the chain via `setPluginManager` in case you want to use methods like `FilterPluginManager#attachByName`. We could remove these as internally it just uses the `PluginManager` to "assemble" the filter.

https://github.com/laminas/laminas-filter/blob/d68ed259498303fedc9a401e91bf828ca41f474e/src/FilterChain.php#L145
https://github.com/laminas/laminas-filter/blob/d68ed259498303fedc9a401e91bf828ca41f474e/src/FilterChain.php#L196

But that might lead to issues when it comes to the whole form attribute stuff like:

```php

use Laminas\Form\Annotation;

final class MyEmailFormular
{
    #[Annotation\Filter(StringTrim::class)]
    #[Annotation\Filter(StripTags::class)]
    #[Annotation\Validator(EmailAddress::class, ['useMxCheck' => true])]
    public string $email;
}
```

So there still needs to be a way to have `AttachByName` available as you do not want validators/filters to be instantiated in the attribute notation (as some validators might be more complex - requiring database connections and stuff).

boesing:
I recently made changes to `PluginManagerInterface#get` which should not return weird types anymore:

https://github.com/laminas/laminas-servicemanager/blob/827be21101110fb062fcd917a047ce9a21cebc3f/src/PluginManagerInterface.php#L32-L34

Can you verify if these changes solve your problem?

boesing:
Closing here. If you think this is not properly covered by v4.0.0, please let me know.

PR #175, opened 2023-02-15
Allow psr/container 2.0.*

tobilektri:
Should allow psr/container 2.0.2 - not only 1.1.2

Changes:
https://github.com/php-fig/container/compare/2.0.2...1.1.2

Can close after merge: https://github.com/laminas/laminas-servicemanager/pull/151

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "laminas/laminas-stdlib": "^3.2.1",
-        "psr/container": "^1.0"
+        "psr/container": "^1.0|^2.0"
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11.99.5",
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -269,7 +269,7 @@ class ServiceManager implements ServiceLocatorInterface
      * @param string|class-string $name
      * @return bool
      */
-    public function has($name)
+    public function has($name): bool
     {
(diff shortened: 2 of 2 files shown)


Ocramius (changes requested):
Sadly not as simple as it sounds, as this represents a major BC break for downstream.

Ocramius:
Duplicate of #151

tobilektri:
@Ocramius could you please explain why adding a new allowed version is a "major BC break for downstream" - the change in the psr/container is not a major BC in my opinion. 
What has to be fixed to allow the ^2.0?

Ocramius:
This snippet makes it more clear: https://3v4l.org/DfZYG

```php
<?php

class ServiceManager {
    function has($name): bool { return true; } 
}

final class UserlandContainer extends ServiceManager
{
    function has($name) { return true; } 
}
```

```
Fatal error: Declaration of UserlandContainer::has($name) must be compatible with ServiceManager::has($name): bool in /in/DfZYG on line 9

Process exited with code 255.
```

tobilektri:
So this change will only take place with the (next?) Major Version release of laminas-servicemanager?

Edit: Thanks for the example! :)

froschdesign:
> So this change will only take place with the (next?) Major Version release of laminas-servicemanager?

Correct. See also the related [milestone for the next major version](https://github.com/laminas/laminas-servicemanager/milestone/2).

PR #176, opened 2023-02-21
Convert old CLI standalone commands to laminas-cli commands

### LINKED ISSUE #172: Convert commands to laminas-cli symfony commands
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Summary

There is `ConfigDumper` command and `FactoryCreator` command available in this component. These commands should be migrated to `laminas-cli` in v4.0.0.

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes

### Description

This converts the existing CLI commands to symfony commands which can be used with `laminas-cli`.
These should get automatically registered to upstream projects once `symfony/console` is available.


froschdesign:
@boesing 
I will take over the documentation. See #177

src/Command/ConfigDumperCommand.php
```diff
@@ -0,0 +1,169 @@
...
+use function class_exists;
+use function dirname;
+use function file_exists;
+use function file_put_contents;
+use function is_array;
+use function is_string;
+use function is_writable;
+use function sprintf;
+
+final class ConfigDumperCommand extends Command
```
Ocramius, line 31: `symfony/console` is in `require-dev`, yet we have a direct dependency to it under `src/` here :thinking:

boesing: Yeah, commands are not meant to be used like services.
I did the same in `laminas-cache` as I do not require upstream projects to have `laminas-cli`.

The `ConfigProvider` does not inject anything `laminas-cli` related until `symfony/console` is available (which would be the case when either a `laminas-cli` or `symfony/console` is used.

Therefore, I've added it as an optional dependency - but might worth to add it to `suggest` 🤔 

The services (which are meant to be re-used) can still be used without `symfony/console` tho.


So I guess we have two options:
1. move `symfony/console` to `require`
2. add `laminas-cli` to `suggest` and mark the commands as `@internal` which implies it should not be instantiated from outside

WDYT?

----

https://github.com/laminas/laminas-servicemanager/pull/176/files#diff-f0a2f25e61573ecaca3a489ef6fca6a6106b7f825ef5599ec374f5387db58f82R49-R54

boesing: Since there was no further feedback, I'll go with option 2 here:

- [x] add laminas-cli to `suggest` and mark the commands as `@internal` which for CLI commands implies it should not be instantiated from outside

froschdesign: Sorry for the late response, but I like the solution as in laminas-cache. 👍🏻

src/ConfigProvider.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * @return array<string,mixed>
+     */
+    private function getLaminasCliDependencies(): array
+    {
+        if (PHP_SAPI !== 'cli') {
+            return [];
+        }
```
Xerkus: This is problematic approach with config since config cache does not distinguish cli and running under any other SAPI.

Commands introduced here are for development where cache should not be enabled generally speaking, but this is the kind of approach that should not be used at all.

src/Command/ConfigDumperCommand.php
Xerkus: For the record, I am fine with taken approach on following conditions:
- Commands are to be used in development only.
- Commands are not meant to be used in any kind of automation.

If it is something that would be used in CI/CD pipeline or in deployments it needs to be via satellite package with explicit dependencies.

Xerkus (changes requested):

src/ConfigProvider.php
froschdesign: @Xerkus 
> Commands introduced here are for development where cache should not be enabled generally speaking…

I would even go one step further: the commands will not work correctly when the cache is active.

> …but this is the kind of approach that should not be used at all.

What do you mean here exactly?

Xerkus: > What do you mean here exactly?

I mean the config provider approach resulting in differing configuration based on some runtime condition. Like environment variable or SAPI type as in this instance. The condition does not have meaningful side effects here but it sets a precedent for the practice I can not recommend. Since it does not have meaningful effect I do not see a reason to use it in the first place.

I mean, application can choose to do it for a good reason but not generic component in its config provider.

boesing: I can follow the reasoning here, esp. when it comes to config caches.
But again, we are talking about CLI-only tooling and it also verifies that another dependency has to be available.
Don't you think that this is okayish even tho it can have side-effects?

I mean, its not that we blindly merge code just because s1 says "but in PR XY some1 already did that as well and now I want my stuff in here as well based on runtime stuff".

We might have to carefully check what is being added but I cannot think of any CLI dependency which is provided by the servicemanager which needs to be available on production environments and thus, config cache should be disabled in those environments where the commands are in use anyways.

WDYT? Do you have better suggestions on how to provide these CLI commands in a more optional way?
i.e. via delegators?

Xerkus: > I mean, its not that we blindly merge code just because s1 says "but in PR XY some1 already did that as well and now I want my stuff in here as well based on runtime stuff".

It is more of "they do it, that is how it is intended to be done" case. I would rather not have to check this at all and not give ideas to people doing third party stuff.

Because this approach in general triggers red flags in my mind I want a compelling reason to have it used rather than just "is okayish".

boesing: As discussed via Slack:

- [x] I'll drop the `cli` SAPI check as everything would work without it as well.

Issue #178, opened 2023-02-21
Create migration guide for version 4


froschdesign:
The following changes must be covered:

- #170 
- #172 
- #179
- _must be continued…_

boesing:
AFAIR, there is a rector for the deprecated factories conversion.
/cc @samsonasik

boesing:
I'll remove this issue from v4 to get this out. We can add migration guide afterwards as well.

dkmuir:
It's been almost a year since 4.0 was released. Is the migration guide being delayed until the rest of Laminas has been upgraded? Or are the other components waiting for the migration guide before upgrading?

froschdesign:
@dkmuir 
Thanks for interest!
At the moment we are busy with introducing the new version in other components. The new major version of laminas-filter is on the way and the next one are laminas-inputfilter, laminas-form and laminas-view.
If you need help with the migration to version 4 of laminas-servicemanager then you can use [our forum or chat](https://getlaminas.org/participate/).

On the other hand, any support is of course welcome! 👍🏻

PR #179, opened 2023-02-23
Remove `ServiceManager` inheritance for `AbstractPluginManager`

### LINKED ISSUE #44: Add PHP 7.3 type declarations

### LINKED ISSUE #89: AbstractPluginManager::get has parameter name mismatch with ContainerInterface::get
Found in https://github.com/laminas/laminas-form/pull/121#discussion_r639546501

As of PHP 8 named parameters, `\Laminas\ServiceManager\AbstractPluginManager::get(name)` should be renamed to `\Laminas\ServiceManager\AbstractPluginManager::get(id)`

### LINKED ISSUE #159: Second param of `ServiceManager::setFactory` doesn't accept a `class-string` of a callable class
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 3.17.0

#### Summary

When calling `\Laminas\ServiceManager\ServiceManager::setFactory` the `@psalm-param` [declaration](https://github.com/laminas/laminas-servicemanager/blob/a54c36e6fd81ba10c94f123e307c4cefd94b1d72/src/ServiceManager.php#L401) of the `$factory` param requires that the `class-string` implements the `\Laminas\ServiceManager\Factory\FactoryInterface`.
The service manager also [accepts a plain `callable`](https://docs.laminas.dev/laminas-servicemanager/configuring-the-service-manager/#factories), though.

#### Current behavior

PHPStan complains that the `class-string`  of the class doesn't implement `FactoryInterface`.

#### How to reproduce

Try to pass a callable class to `setFactory` and run PHPStan.

#### Expected behavior

The method should accept a plain callable class. Psalm and PHPStan don't accept `class-string<callable>`, though.

### LINKED ISSUE #174: Mark `ServiceManager` as `final` and change `AbstractPluginManager` to use composition over inheritance
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 4.0.0  |
| BC Break?           | Yes |

### Goal

Remove inheritance of `ServiceManager`.

### Background

Besides the fact that composition over inheritance should be preferred, the `AbstractPluginManager` is overriding methods from `ServiceManager` which should be delegated instead.

### Considerations

Upstream projects might type against `PluginManagerInterface` or the specific plugin manager instance instead of typing against `ServiceManager` and thus the BC break would be not too big.

### Proposal(s)

- create clean `ServiceManager` instance within `__construct` and configure that service manager with the config for the plugin manager.
- remove  `ServiceManager` specific public methods
- only support `get`, `has`, `build` as declared within `PluginManagerInterface`
- create `ConfigurableInterface` which provides `configure` method to consume `ServiceManagerConfigurationType` which passes the argument to the pluginmanager related `ServiceManager`
- make `PluginManagerInterface` implement `ConfigurableInterface` so that all plugin managers have to implement that interface(s) method

### LINKED ISSUE #182: Remove `AbstractPluginManager#validate`
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 4.0.0  |
| BC Break?           | Yes |

### Goal

Due to the current implementation, implementors of the `AbstractPluginManager` can easily "oversee" that they have to provide `$instanceOf` property to provide any validation of their plugin manager. This leads to implementations with no validation at all:

https://psalm.dev/r/d12af359e2

### Background

Even tho, the current implementation within `AbstractPluginManager` also allows adding complex generics to their plugin manager without adding any validation. Might be even the case that implementors think that laminas got their back do the fact that there is already an `AbstractPluginManager#validate`.

Given the example of https://psalm.dev/r/d12af359e2, it is quite easy to implement a `PluginManager` which can return `mixed` while stating to only return `int|string`.

### Considerations

Even tho, that in `laminas-servicemanager` v3 the `validatePlugin` method was removed (which had to be implemented in implementing classes) - for convenience - I'd require it to be implemented again with v4 while providing a trait which can be implemented in case that a `PluginManager` did not need more than that.
In order to have more type-stability, this change would be mandatory tho.

### Proposal(s)

- Remove `AbstractPluginManager#validate` and `AbstractPluginManager#$instanceOf`
- Provide an `SingleInstanceOfPluginManagerTrait` which provides that functionality so it can be re-implemented conveniently

### Appendix

```php
/** 
 * @template InstanceType of object
 * @psalm-require-implements PluginManagerInterface
 */
trait Foo
{
    /** @var class-string<InstanceType> */
    protected string $instanceOf;
    
    /** @psalm-assert InstanceType $instance */
    public function validate(mixed $instance): void
    {
        if($instance instanceOf $this->instanceOf) {
            return;
        }
        throw new InvalidArgumentException('Not the expected instance of...');
    }
}

/** 
 * @implements PluginManagerInterface<stdClass> 
 */
final class Bar implements PluginManagerInterface
{
    /** @use Foo<stdClass> */
    use Foo;
    
    protected string $instanceOf = stdClass::class;
}
```

https://psalm.dev/r/9a75a42fda

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| BC Break      | yes

### Description

This is a PoC of how the plugin manager could work without inheriting from the `ServiceManager`.
With this PoC, I've also modified `ServiceManager` to return `mixed` as thats actually the case already. Nothing in this component verifies that something is `object`. We do have soft-types for factories to only return `object` but in fact, they can return `mixed` anyways.
Since the `AbstractPluginManager` does not require an `$instanceOf` value and thus, can also return `mixed` in case it is not set, I think this change makes sense until we decide to go the same route as [symfony](https://github.com/symfony/dependency-injection/blob/83369dd4ec84bba9673524d25b79dfbde9e6e84c/ContainerInterface.php#L41) does. Since that decision has not yet been made, I'd prefer to skip that. I'm open to create an RFC for this and bring it to the next TSC if some1 wants to further discuss this. 

Please NOTE that I also:

- removed some useless tests which verified exceptions for passing invalid value sto either `ServiceManager` and/or `AbstractPluginManager`
- added plenty of type-hints (even to interfaces) - but not to every interface (yet) - will definitely do that before finally creating a v4 (@samsonasik would you mind updating the `laminas/laminas-servicemanager-migration` rector rules once we finished with adding all the missing types?)
- slightly modified the `ServiceManagerConfigurationType` (still compatible but more strict)
- with a psalm version containing [this patch](https://github.com/vimeo/psalm/pull/9599), we can provide even better psalm-types regarding callable-objects and class-string

TODOS:

- [x] Modify container-config-test repository so that some tests are being dropped
- [ ] Mark `ConfigInterface` and `Config` as deprecated in v3
- [ ] Add psalm-types from `ConfigInterface` to `ServiceManager` in a redundant way (in v3) so that upstream projects can already switch 
- [ ] Mark `ServiceManager` as `@final` in v3
- [x] Get https://github.com/vimeo/psalm/pull/9599 merged so that we can provide even better psalm-types for `class-string<callable(ContainerInterface,string,array|null):mixed>`


test/CommonServiceLocatorBehaviorsTrait.php
```diff
@@ -734,25 +673,6 @@ public function testCanInjectFactories(): void
         self::assertSame($instance, $foo);
     }
 
-    /**
-     * @group mutation
-     * @covers \Laminas\ServiceManager\ServiceManager::mapLazyService
-     */
-    public function testCanMapLazyServices(): void
```
boesing, line 737: Moved this test to the `ServiceManager` specific tests as this is used for both `AbstractPluginManager` AND `ServiceManager`.

I could've checked the container instance and marked as skipped but since there is already a dedicated test class, I moved it.

test/ConfigTest.php
```diff
@@ -4,86 +4,51 @@
...
+use LaminasTest\ServiceManager\TestAsset\SimpleInitializer;
 use PHPUnit\Framework\TestCase;
 
 /**
  * @covers \Laminas\ServiceManager\Config
+ * @psalm-import-type ServiceManagerConfigurationType from ConfigInterface
  */
 final class ConfigTest extends TestCase
 {
-    public function testMergeArrays(): void
```
boesing, line 16: removed this feature. The `Config` class is marked `final` now, I have no idea why some1 needs a dedicated config for a plugin manager. If still needed, creating a decorator to re-implement this feature in upstream projects seems to be good enough for me.

/cc @froschdesign I guess that needs to be added to the migration guide?

samsonasik:
@boesing ok , I can do that 👍 , I may need summary which part that need update

test/ConfigTest.php
froschdesign: I added this pull request to the related issue report: #178

boesing:
Partly solves #44

psalm.xml.dist
```diff
@@ -47,6 +49,19 @@
...
+            <errorLevel type="info">
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::addAbstractFactory"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::setService"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::setInvokableClass"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::setAlias"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::setShared"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::addInitializer"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::addDelegator"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::setFactory"/>
+                <referencedMethod name="Laminas\ServiceManager\AbstractPluginManager::mapLazyService"/>
```
Ocramius, line 70: Should we get rid of these completely, or at least make them `private`? :thinking:

src/AbstractFactory/ReflectionBasedAbstractFactory.php
```diff
@@ -105,20 +107,25 @@ public function __construct(array $aliases = [])
...
      * {@inheritDoc}
-     *
-     * @return DispatchableInterface
      */
-    public function __invoke(ContainerInterface $container, $requestedName, ?array $options = null)
+    public function __invoke(ContainerInterface $container, string $requestedName, ?array $options = null): mixed
     {
+        if (! class_exists($requestedName)) {
+            throw new InvalidArgumentException(sprintf('%s can only be used with class names.', self::class));
+        }
```
Ocramius: IMO this runtime check can be removed, replaced with:

 1. an `@param class-string $requestedName`
 2. a suppression (we obviously aren't honoring the parent type)

`new ReflectionClass($requestedName)` already performs this check for us anyway.

src/AbstractPluginManager.php
```diff
@@ -201,21 +139,143 @@ public function validate(mixed $instance)
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function build(string $name, ?array $options = null): mixed
+    {
+        /** @psalm-suppress MixedAssignment Yes indeed, service managers can return mixed. */
+        $plugin = $this->plugins->build($name, $options);
+        $this->validate($plugin);
```
Ocramius, line 221: Discussed in #182

I overall think we need to redesign this class, but I don't mind doing it in a patch addressing #182 directly.

psalm.xml.dist
Ocramius: Never mind: seen the deprecations below :)

src/Config.php
```diff
@@ -25,10 +21,9 @@
  *
  * @psalm-import-type ServiceManagerConfigurationType from ConfigInterface
  */
-class Config implements ConfigInterface
+final class Config implements ConfigInterface
 {
-    /** @var array<string,bool> */
-    private array $allowedKeys = [
+    private const ALLOWED_CONFIGURATION_KEYS = [
```
Ocramius: This constant is now unused

src/Config.php
```diff
@@ -40,63 +35,30 @@ class Config implements ConfigInterface
...
-        'delegators'         => [],
-        'factories'          => [],
-        'initializers'       => [],
-        'invokables'         => [],
-        'lazy_services'      => [],
-        'services'           => [],
-        'shared'             => [],
-    ];
+    /** @var ServiceManagerConfigurationType */
+    private array $config;
```
Ocramius: Not to be tackled here, but we may (in future) benchmark the creation and usage of this object with a nested array vs declared properties.

src/PluginManagerInterface.php
```diff
@@ -13,17 +15,53 @@
...
+
+    /**
+     * @param ServiceManagerConfigurationType $config
+     * @throws ContainerModificationsNotAllowedException If the allow override flag has been toggled off, and a
+     *                                                   service instanceexists for a given service.
+     * @throws InvalidServiceException If an instance passed in the `services` configuration is invalid for the
+     *                                 plugin manager.
+     * @throws CyclicAliasException If the configuration contains aliases targeting themselves.
+     */
+    public function configure(array $config): static;
```
Ocramius: IMO `configure()` shouldn't be part of the public API of a `PluginManagerInterface`, since the plugin manager is likely going to become a thin wrapper around a more complex and potentially completely immutable container.

Was it introduced on purpose here? What are the implication if we drop this?

src/ServiceLocatorInterface.php
```diff
@@ -17,14 +23,20 @@ interface ServiceLocatorInterface extends ContainerInterface
...
-    public function build($name, ?array $options = null);
+    public function build(string $name, ?array $options = null): mixed;
+
+    /**
+     * @param ServiceManagerConfigurationType $config
+     * @throws ContainerModificationsNotAllowedException If the allow override flag has been toggled off, and a
+     *                                                   service instanceexists for a given service.
+     * @throws CyclicAliasException If the configuration contains aliases targeting themselves.
+     */
+    public function configure(array $config): static;
```
Ocramius: Similar to `PluginManagerInterface`: this should absolutely not be here.

src/ServiceManager.php
```diff
@@ -59,19 +61,12 @@
...
- *     factories?: FactoriesConfigurationType,
- *     initializers?: InitializersConfigurationType,
- *     invokables?: array<string,string>,
- *     lazy_services?: LazyServicesConfigurationType,
- *     services?: array<string,object|array>,
- *     shared?:array<string,bool>,
- *     shared_by_default?:bool,
- *     ...
- * }
+ * @psalm-import-type ServiceManagerConfigurationType from ConfigInterface
```
Ocramius: Not sure if we should alias `ServiceManagerConfiguration` here, mostly to ease downstream usage?

src/ServiceManager.php
```diff
@@ -239,27 +229,24 @@ public function get($name)
...
-            $this->services[$resolvedName] = $object;
+            $this->services[$resolvedName] = $service;
+            $this->services[$name]         = $service;
         }
 
-        // Also cache under the alias name; this allows sharing based on the
-        // service name used.
-        if ($sharedAlias) {
-            $this->services[$name] = $object;
-        }
```
Ocramius, line 231: Removed on purpose? :thinking: 

Alias caching was kinda relevant for performance

src/ServiceManager.php
```diff
@@ -976,13 +939,31 @@ private function assertCallableDelegatorFactory($delegatorFactory): void
...
             is_object($delegatorFactory) ? $delegatorFactory::class : gettype($delegatorFactory),
-            Factory\DelegatorFactoryInterface::class
+            DelegatorFactoryInterface::class
         ));
     }
+
+    /**
+     * @return DelegatorFactoryInterface|DelegatorCallableType
+     */
+    private function resolveDelegatorFactory(mixed $delegatorFactory): callable|DelegatorFactoryInterface
```
Ocramius: Assuming well-formed configuration, can we perhaps be stricter on the input type here?

Ocramius (changes requested):
My only major problem with this patch is the addition of `#configure()` on interfaces that are designed for container consumption, rather than configuration.

Configuration of a container can happen on concrete instances, while the `ContainerInterface`, `ServiceLocatorInterface` and `PluginManagerInterface` interfaces should be read-only, without any assumption that they might be configurable.

Heck, they may be dumped/generated code.

src/PluginManagerInterface.php
boesing: The various cache adapters are adding themselves to the CacheAdapterPluginManager via delegator factories. Since every plugin manager was extending servicemanager, that method was available before even tho it was nowhere declared. But maybe we can drop it from the Interface as usually the concrete pluginmanager is being fetched anyways.

Ocramius: Yeah, I'd prefer typing `ServiceManager` there, instead of typing against the interface then :+1:

src/ServiceManager.php
boesing: There is a check if it is a shared alias a few lines above and thus the if was redundant. Just moved the caching to another if which should suffice. 

L238

src/PluginManagerInterface.php
boesing: But the delegators are not able to Access the SM directly. Due to the removal of the inheritance, there is no way to modify the SM from the delegators - thats why these methods were kept.

src/ServiceManager.php
Ocramius: Ah, I see that indeed you added some code around `if ($sharedService) {` :+1:

Xerkus:
`configure()` is/was part of mutable container. I remember there was a push to make container immutable and configurable at creation. It got ultimately dropped unmerged or reverted before release. 

Its use case might have been related to how container or various plugin managers were configured after creation in module manager/mvc events.

I am for making it internal implementation detail and removing from public interface.

src/PluginManagerInterface.php
Ocramius: Even then, I'd say that this is a design smell: you cannot assume that the container can be modified post-instantiation, while this method addition makes that a requirement of any implementation.

Xerkus:
`configure()` is indeed mvc late configuration related. Context could be found here  https://github.com/zendframework/zend-servicemanager/pull/47

boesing:
@Xerkus @Ocramius I'd be up for a read-only container with v5. Given the fact that v4 to the end of Q1 is already tough with all the migration paths, I'd prefer not being too complex here. 

Keeping the old methods is mostly due to BC compat but I am fine with dropping configure from the Interface. WDYT?

Ocramius:
Dropping from the interface(s) is sufficient for me, for now :+1:

src/PluginManagerInterface.php
boesing: Yah, I've removed the method from the interface while keeping it within the `AbstractPluginManager`. 

Are we fine with creating a frozen `ServiceManager` concept for the next major version rather than adding it to v4? 
v4 is overdue for years now and it feels like something no1 prepared in the last years and thus, at least for now, it should be fine with going the current direction. This will also avoid having too many refactoring requirements to consumers for this version.

src/ServiceLocatorInterface.php
boesing: Removed.

src/ServiceManager.php
boesing: Thats the import, has to be modified in the `ConfigInterface` tho.

One of the problems I do have with not having the `*Type` suffix is, that it looks in the rest of the file as it would be an interface or a class. Having the `Type`-suffix notation (I sometimes also prefer `ArrayShape`) makes it clear, that this is some sort of custom type.

So mainly I choose this for readability. But if you say that it is okay to remove it, I dont have a too strong opinion on this and open for changes.

src/PluginManagerInterface.php
Xerkus: I do not think immutability should be dropped in the mix here. Make the move to PSR container as painless as it can be first.

src/Config.php
boesing: I mark this as deprecated, as the object makes no sense anymore. 
I also dont think that an object will be more performant than an assoc array.

src/AbstractFactory/ReflectionBasedAbstractFactory.php
boesing: This would open a huge can of worms:

1. `FactoryInterface` does not state `$requestedName` to be `class-string` as factories can return whatever
2. If I annotate `class-string` for the `ReflectionBasedAbstractFactory`, I also would've to suppress `MoreSpecificImplementedParamType`
3. After I've done that, I can remove the `class_exists` check

But for what? All factories are called from within the `ServiceManager`, there is no static analysis (yet) which prevents a user to actually use the factory with interfaces (even tho, that will lead to runtime errors).

So I'd rather keep the runtime check and provide a proper error rather than adding suppressions.

src/Config.php
boesing: Dropped the whole file, we are targeting next major anyways and thus can be dropped.
Have to mark it as deprecated in v3 after merging this PR tho.

src/ServiceManager.php
boesing: Thought about it again, it is `ServiceManagerConfigurationType` as of now but since upstream projects still have to modify the usage from `ConfigInterface` to `ServiceManager`, it should be possible to rename this as well.

- [x] gonna rename this

src/ServiceManager.php
boesing: Done that in the recent commits.

test/ConfigTest.php
boesing: I totally removed `Config` + `ConfigInterface` in my latest work. Its just a useless class/interface anyways since the config could've always been passed to `ServiceManager#__construct`.

boesing:
- [x] Check why psalm runs into endless loop, had these errors locally as well but mostly after clear-cache that error has gone. Seems not to be gone in GHA even tho there is no cache...

src/AbstractPluginManager.php
```diff
@@ -5,217 +5,345 @@
...
+ * The constructor accepts the parent container instance, which is then used when creating instances.
  *
  * @template InstanceType
- * @implements PluginManagerInterface<InstanceType>
+ * @template-implements PluginManagerInterface<InstanceType>
  * @psalm-import-type ServiceManagerConfiguration from ServiceManager
- * @psalm-suppress PropertyNotSetInConstructor
+ * @psalm-import-type FactoryCallable from ServiceManager
+ * @psalm-import-type DelegatorCallable from ServiceManager
+ * @psalm-import-type InitializerCallable from ServiceManager
```
Xerkus, line 31: Removing all 3 callable imports breaks circular reference somewhere I presume, avoiding the psalm erroring with OOM.

`/usr/bin/time -f '%M' composer static-analysis -- --no-cache`
Gives 412992 in kbytes.

composer.json
```diff
@@ -48,16 +47,15 @@
         "composer/package-versions-deprecated": "^1.11.99.5",
         "laminas/laminas-cli": "^1.8",
         "laminas/laminas-coding-standard": "~2.5.0",
-        "laminas/laminas-container-config-test": "^0.8",
-        "laminas/laminas-dependency-plugin": "^2.2",
+        "laminas/laminas-container-config-test": "dev-qa/service-manager-v4",
         "lctrs/psalm-psr-container-plugin": "^1.9",
         "mikey179/vfsstream": "^1.6.11@alpha",
-        "ocramius/proxy-manager": "^2.14.1",
+        "friendsofphp/proxy-manager-lts": "^1",
```
boesing, line 53: Should be extracted to a dedicated satellite package which decorates the container to add proxy functionality.
Is that something which can be easily implemented @Ocramius?

composer.json
```diff
@@ -48,16 +47,15 @@
...
         "phpbench/phpbench": "^1.2.7",
         "phpunit/phpunit": "^9.5.26",
         "psalm/plugin-phpunit": "^0.18.0",
         "symfony/console": "^6.0",
-        "vimeo/psalm": "^5.0.0"
+        "vimeo/psalm": "5.x-dev"
```
boesing: - [ ] needs to be switched back once psalm releases the next minor

composer.json
boesing: https://github.com/laminas/laminas-servicemanager/issues/186

src/PluginManagerInterface.php
```diff
@@ -19,11 +19,38 @@ interface PluginManagerInterface extends ServiceLocatorInterface
...
+     * @template TRequestedInstance extends InstanceType
+     * @psalm-param class-string<TRequestedInstance>|string $id Service name of plugin to retrieve.
+     * @psalm-return ($id is class-string ? TRequestedInstance : InstanceType)
+     * @throws Exception\ServiceNotFoundException If the manager does not have
+     *     a service definition for the instance, and the service is not
+     *     auto-invokable.
+     * @throws InvalidServiceException If the plugin created is invalid for the
+     *     plugin context.
+     */
+    public function get(string $id): mixed;
```
weierophinney: Since `ServiceLocatorInterface` already extends `ContainerInterface`, should these really be defined here?

I can understand defining `get()` so we can add the additional annotations, but at the very least, `has()` feels superfluous.

weierophinney (commented):
Overall,I like this approach. Composing a `ServiceManager` inside the `AbstractPluginManager` removes a lot of duplication, while simultaneously removing the inheritance. Having an interface means we can use that as the type instead of an abstract class, which also makes me happy. And the psalm shapes in the `ServiceManager` provide valuable documentation of expectations.

My only concern/issue is that I don't recall the the rationale for deprecating/removing the `ConfigInterface` and `Config` class. I recognize that we can use Psalm annotations to provide shapes for valid data, but that seems more likely to lead to issues at runtime where the shapes cannot be verified/validated. That said, I won't block approval on that; just wish I could remember what the arguments were around this.

src/PluginManagerInterface.php
boesing: Yah, `has` is superfluous indeed. 
I guess I've added it to also add some `psalm-assert` on it to provide information of what will be returned from `get`. But since `has` actually does no validation at all, it would provide false information and thus we can safely drop it here.

boesing:
>  [...] rationale for deprecating/removing the ConfigInterface and Config [...]

I don't see any value for that interface or the Config. The interface itself was not used anywhere and the `Config` was almost always empty.
We (in my company) configured our project container like this in mezzio:

__container.php__
```php
<?php

use Laminas\ServiceManager\Config;
use Laminas\ServiceManager\ServiceManager;

require_once __DIR__ . '/../vendor/autoload.php';

// Load configuration
$config = require __DIR__ . '/config.php';

// Build container
$container = new ServiceManager();
(new Config($config['dependencies']))->configureServiceManager($container);

// Inject config
$container->setService('Config', $config);

// Make container writable since we want to overwrite even existing alias for config
$container->setAllowOverride(true);
$container->setAlias('config', 'Config');
$container->setAllowOverride(false);

return $container;
```

The laminas-modulemanager seems to provide support for the `ConfigInterface` and maybe some projects out there might have used that. I guess that was to avoid having to provide the array-structure from `Module#getConfig`. But since that is the usual way of how it is returned from `ConfigProvider#__invoke` in mezzio applications, the plan for laminas-mvc v4 is to use the config-aggregator for modules as well (hopefully, that would be huge, can be even done as of now by using [laminas/laminas-config-aggregator-modulemanager](https://github.com/laminas/laminas-config-aggregator-modulemanager)) and due to the fact that we have psalm-types for a period of time now gives me the feeling that the `ConfigInterface` along with its implementation is adding unnecessary code-paths to applications (as in the end, we end up having an array anyways which is passed to `ServiceManager#configure` or `AbstractPluginManager#configure`.

So its just a dangling class mostly used to have a value-object of the config which wasnt properly validated at runtime and didn't provide any type until [3.9.0](https://github.com/laminas/laminas-servicemanager/releases/tag/3.9.0) which was 11 years after the initial creation of the interface.

----

Thats why I think we can safely drop it and let upstream projects actually return the array structure while using psalm + psalm-type to have proper type validation:

__before__

```php
<?php

namespace MyModule;

use Laminas\ModuleManager\Feature\ServiceProviderInterface;
use Laminas\ServiceManager\Config;
use Laminas\ServiceManager\Factory\InvokableFactory;

final class Module implements ServiceProviderInterface
{

    public function getServiceConfig(): Config
    {
        return new Config([
            'factories' => [
                MyModuleService::class => InvokableFactory::class,
            ],
        ]);
    }
}
```

__after__
```php
<?php

namespace MyModule;

use Laminas\ModuleManager\Feature\ServiceProviderInterface;
use Laminas\ServiceManager\Factory\InvokableFactory;
use Laminas\ServiceManager\ServiceManager;

/** @psalm-import-type ServiceManagerConfiguration from ServiceManager */
final class Module implements ServiceProviderInterface
{

    /**
     * @return ServiceManagerConfiguration
     */
    public function getServiceConfig(): array
    {
        return [
            'factories' => [
                MyModuleService::class => InvokableFactory::class,
            ],
        ];
    }
}
```

----

However, I am open to keep the `Config` object if we prefer object > array but then I would actually prefer the config being injected as an object to the SM/PluginManager instead of allowing arrays. But then we have `ConfigInterface#toArray` again which does not really change anything.

Should we create an RFC for this? I can restore the `ConfigInterface` + `Config` here for now.
But we would have to rename `ConfigInterface#configureServiceManager` to `ConfigInterface#configure` and then allow both `ServiceManager` + `AbstractPluginManager` due to the removal of inheritance here.

weierophinney:
@boesing Thanks for that summary!

When I consider that the majority of the time, we're passing an array to either the `ServiceManager` or a plugin manager, and that it has to do validations internally _anyways_ due to this, it doesn't make sense to keep it. The only reason to keep it is if the SM and/or PM _required_ an instance, and pulled configuration from it, as it would then provide validations. That would just be shifting responsibilities, and would introduce duplications regardless, as we have the various `set*()` methods, and they have to do validations, too.

boesing:
> @boesing ok , I can do that 👍 , I may need summary which part that need update

@samsonasik I guess I've finished the work here, so I can provide a lil summary:

1. all classes extending `AbstractPluginManager` should either 
1.1 extend `AbstractSingleInstancePluginManager` in case the `$instanceOf` property is declared and should adapt `$instanceOf` value in `@template-extends` annotation
1.2 implement its own `validate` method (can be created as an empty method from rector in case it is missing)
2. We could search for classes implementing the [modulemanager](https://github.com/laminas/laminas-modulemanager/blob/650a98b11cdd676c722a8925187b288da5808e34/src/Feature/ServiceProviderInterface.php#L17) `ServiceProviderInterface#getServiceConfig` and in case of a return value of `Config` it should return `array` (basically the `__construct($arg)` which is being passed to the `Config`. Refer to https://github.com/laminas/laminas-servicemanager/pull/179#issuecomment-1525999402 to see how it needs to be replaced.
3. adding property types of `AbstractPluginManager`/ `AbstractSingleInstancePluginManager` (not sure if rector can modify code which is not valid as per linting due to mismatching property types?)
4. `container.php` as in https://github.com/laminas/laminas-servicemanager/pull/179#issuecomment-1525999402 can be modified to `$container = new ServiceManager($config['dependencies']);` instead


Not 100% sure if that is already everything but I guess at least for this PR that should reflect almost all changes.

samsonasik:
@boesing Thank you for the summary, I will try what I can do 👍

weierophinney (approved):
:ship:

PR #180, opened 2023-02-24
Ahead of time factory creation command

### LINKED ISSUE #171: AOT factory creation CLI command for `ReflectionBasedAbstractFactory` mapped factories
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | kinda
| BC Break    | no

#### Summary

Currently, there are CLI commands around where it is possible to create mezzio handler factories and stuff by using reflection. I was wondering if it would also be possible to provide a CLI command which takes runtime `config` from container and creates factories (along with a mapping-file such as `mezzio-tools`) using the same logic as `ReflectionBasedAbstractFactory`.

This would enable projects to use a more "generic" runtime factory during development while enabling devs to run that AoT CLI command in CI builds to auto-generate factories. This will remove the reflection stuff during runtime and increases performance with just running a CLI command (and maybe a bit of configuration, i.e. to provide different other PluginManager configuration keys such as `validators`, etc. - which can also be injected to the config from within our own components).

WDYT?

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| New Feature   | yes
| RFC           | yes

### Description

Fixes #171

So just to quickly sum this up:

We do have project configuration, so this is not some fictional use-case.
This configuration contains `service_manager` configuration where some services (usually only depending on interfaces) are registered via `ReflectionBasedAbstractFactory`:

```php
return [
    'service_manager' => [
        'factories' => [
            OurService::class => ReflectionBasedAbstractFactory::class,
        ],
    ],
];
```

Even tho, the `ReflectionBasedAbstractFactory` is an abstract factory, it also extends `FactoryInterface` and thus can be used as shown above. 

- One of the "good" things is, that there is no need for a dedicated factory.
- The "bad" thing is, that the factory consumes runtime to use reflection and to determine the appropriate `__construct` argument values

I've played a little bit in our project and I ended up writing some code (which I now used to contribute here) which allows us to use `ReflectionBasedAbstractFactory` in development while providing a CLI command to generate a opcachable factory while putting it on the filesystem during release artifact generation.

Usage would be as follows:

```plain
Description:
  Creates factories which replace the runtime overhead for `ReflectionBasedAbstractFactory`.

Usage:
  servicemanager:generate-aot-factories [<localConfigFilename>]

Arguments:
  localConfigFilename        Should be a path targeting a filename which will be created so that the config autoloading will pick it up. Using a `.local.php` suffix should verify that the file is overriding existing configuration. [default: "config/autoload/generated-factories.local.php"]
```

One needs also to register a directory in the application config:

**config/autoload/global.php**
```php
return [
    \Laminas\ServiceManager\ConfigProvider::CONFIGURATION_KEY_FACTORY_TARGET_PATH => 'path/to/config/autoloading/directory',
];
```

Since the path registered in that configuration key MUST also be added to `composer.json` `autoload` section using `classmap`, having this as a required configuration key to use the CLI command seemed to be a good solution.

```json
{
     "autoload": {"classmap": ["path/to/factory/output"]}
}
```

So after configuring the application as mentioned above, the CLI command could be executed like:
```bash
$ php vendor/bin/laminas servicemanager:generate-aot-factories config/autoload/aot-factories.local.php
```

This will then scan the projects configuration for **all** container configurations (so everything which looks like a container configuration actually: has to be an array, containing key `factories` being an array, containing values with either `class-string<ReflectionBasedAbstractFactory> `or  instance of `ReflectionBasedAbstractFactory`.

So in the end, we should have a bunch of generated factories located in the `path/to/factory/output` directory (there is a sub-directory per "container") PLUS the `aot-factories.local.php` file located in `config/autoload`. 

In the end, running `composer dump-autoload` should "register" all the factories located in `path/to/factory/output` and the next `vendor/bin/laminas servicemanager:generate-aot-factories` run should tell us, that there are no (more) services registered via `ReflectionBasedAbstractFactory` (as the `aot-factories.local.php` should've overriden the initial application config).

🎉


boesing:
Actual commit to review:
https://github.com/laminas/laminas-servicemanager/pull/180/commits/bd6a3626a2a5876dfc7c1c7e39ac4a69be2e4991

src/Tool/ConstructorParameterResolverInterface.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\ServiceManager\Tool;
+
+use Psr\Container\ContainerInterface;
+
+interface ConstructorParameterResolverInterface
```
froschdesign: Does this mean that more resolvers will be added in the future? (I am thinking here of forms, for example.)

boesing: Actually not, no. This is just meant to be used with services, IMHO more complex services with scalar argunemt constructors or with a "injection strategy" should stay as is. 

At least I am not planning of supporting more than services available via the servicemanager. 
So anything pluginmanager related wont work with reflection as of now as well. 
But at least having that interface will open up 3rd party to implement more complex resolving strategies. But that wont work with the `ReflectionBasedAbstractFactory` which then requires projects to generate the factories at any time before running the code - thats usually pain in development tho.

froschdesign: > But at least having that interface will open up 3rd party to implement more complex resolving strategies.

👍🏻

bcremer:
I like that proposed feature pretty much :+1: 

But I prefer to keep generated files in a dedicated directory (`./data/generated`).
I can archive that with your proposed feature by adding a new `PhpFileProvider`.

```php
$generatedFactoriesConfig = [
    ServiceManagerConfigProvider::CONFIGURATION_KEY_FACTORY_TARGET_PATH => __DIR__ . '/../../data/generated/',
    'factory-override-path' => __DIR__ . '/../data/generated/generated-factories.php'
];

$aggregator = new ConfigAggregator([
    ServiceManagerConfigProvider::class,

    new PhpFileProvider(realpath(__DIR__) . '/autoload/{,*.}{global,local}.php'),

    // Add config for generated factories
    new ArrayProvider($generatedFactoriesConfig),

    // Add generated factories
    new PhpFileProvider($generatedFactoriesConfig['factory-override-path']),
]);

return $aggregator->getMergedConfig();
```

One feature I'd like to propose is adding `localConfigFilename` to the configuration so I don't have to provide it in the `generate-aot-factories` call.

```bash
php bin/console.php servicemanager:generate-aot-factories  data/generated/generated-factories.php
```

How about `ServiceManagerConfigProvider::CONFIGURATION_KEY_FACTORY_LOCAL_CONFIG_FILENAME`?

WDYT?

boesing:
@bcremer thats exactly how it is proposed here. You have to configure `ConfigProvider:: CONFIGURATION_KEY_FACTORY_TARGET_PATH` to target to a path where the files have to be stored.
If that is missing, executing the command will output that the config key is missing.

You have to add it as your `composer.json` needs that directory in its `autoload` section.

If we are talking about the path which has to be added by `arg1` via CLI - that won't make sense to make it configurable as it could differ from CLI. Not sure if I'd like to have that configurable as well, especially because if it is configurable, who do you expect that file to "add" to your config then? The whole config-merge has to be done via either `ConfigAggregator` or the `moduleloader` and frankly, having runtime logic which requires configs and merge them on-top of the already merged configs is nothing I'd like to support.

So imho, its up to the projects to specify a path and - in case the path is not automatically picked-up by configuration merge - make the changes to their `ConfigAggregator`/`modulemanager` implementations to ensure the file is properly loaded.
So IMHO, you can do what you provided above while synchronizing that path with your CLI execution as you still will have to provide that path via CLI argument to the command.

bcremer:
I withdraw my idea of having a configuration key to configure the path for the override factory mapping (`config/autoload/generated-factories.local.php`).

For most users the the proposed default is the better choice as there is already the default `PhpFileProvider` in place to pick up that file.

I still support the proposed Pull Request from @boesing and would like to see it going forward.

docs/book/ahead-of-time-factories.md
```diff
@@ -0,0 +1,105 @@
+# Ahead of Time Factories 
+
+- Since 4.0.0
+
+In addition to the already existing [Reflection Factory](TODO), one can create factories for those services using `ReflectionBasedAbstractFactory` before deploying the project to production.
```
boesing: @froschdesign How can I link from here to the documentation page of `Reflection Factory`?

froschdesign: The same folder, then:

```markdown
[Reflection Factory](reflection-abstract-factory.md)
```

Reference: https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages

docs/book/ahead-of-time-factories.md
```diff
@@ -0,0 +1,106 @@
+# Ahead of Time Factories
```
froschdesign: We need versioned documentation, so this must be placed in a subdirectory: `docs/book/v4/ahead-of-time-factories.md`
Create the folder and add the file, I will do the rest.

boesing: Done. I duplicated all files and created both `v3` and `v4` folders.

docs/book/v4/ahead-of-time-factories.md
```diff
@@ -0,0 +1,106 @@
...
+
+When the CLI command has finished, there are all factories generated within the path (`ConfigProvider::CONFIGURATION_KEY_FACTORY_TARGET_PATH`) registered in the projects configuration along with the `<target for generated factory config>` file (defaults to `config/autoload/generated-factories.local.php`). It is required to run `composer dump-autoload` (in case you've used optimized/classmap-authoritative flag, you should pass these here again) after executing the CLI command as the autoloader has to pick up the generated factory classes. In case of an existing config cache, it is also mandatory to remove that cached configuration file.
+
+When the project is executed having all the files in-place, the generated factory classes are picked up instead of the `ReflectionBasedAbstractFactory` and thus, no additional runtime side-effects based on `Reflection` will occur.
+
+Ensure that both `<target for generated factory config>` file and the directory (including sub-directories and files) configured within `ConfigProvider::CONFIGURATION_KEY_FACTORY_TARGET_PATH`` is being picked up when generating the artifact which is deployed to production.
```
bcremer: A superfluous back tick is breaking the formatting here:

```diff
- `ConfigProvider::CONFIGURATION_KEY_FACTORY_TARGET_PATH``
+ `ConfigProvider::CONFIGURATION_KEY_FACTORY_TARGET_PATH`
```

docs/book/ahead-of-time-factories.md
froschdesign: Great, I will do the rest.

froschdesign (approved):
The documentation is good, I will take over all further changes, because it would go beyond the scope here.

PR #181, opened 2023-03-05
Remove deprecated `ServiceManager#getServiceLocator` method

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

Removes deprecated `ServiceManager#getServiceLocator` method.


boesing:
I did not remove any deprecation (i.e. those from the `AbstractPluginManager`) since there is #179 which modifies plenty of stuff and also drops deprecations.

PR #183, opened 2023-03-21
Drop PHP 8.0 support, replace `ocramius/proxy-manager` with `friendsofphp/proxy-manager-lts`

Slamdunk:
My hot take on this one.

Unfortunately, https://github.com/Ocramius/ProxyManager hasn't received any attention for PHP after 8.0, and this is blocking `laminas-servicemanager` from being properly upgraded.

This PR is required for https://github.com/laminas/laminas-servicemanager/pull/173, which in turn is required for https://github.com/laminas/laminas-mail/pull/233 any many similar others.

No https://github.com/laminas/laminas-servicemanager/labels/BC%20Break here:

* people who rely on `ocramius/proxy-manager` are still stuck on PHP 8.0 and are going to be stuck on `laminas/laminas-servicemanager:v3.20` forever.
* people who upgraded to PHP >= 8.1 already had to drop `ocramius/proxy-manager` usage because it never added support for newer PHP versions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,72 +13,78 @@
     ],
     "homepage": "https://laminas.dev",
     "support": {
-        "docs": "https://docs.laminas.dev/laminas-servicemanager/",
         "issues": "https://github.com/laminas/laminas-servicemanager/issues",
-        "source": "https://github.com/laminas/laminas-servicemanager",
-        "rss": "https://github.com/laminas/laminas-servicemanager/releases.atom",
+        "forum": "https://discourse.laminas.dev",
         "chat": "https://laminas.dev/chat",
-        "forum": "https://discourse.laminas.dev"
-    },
-    "config": {
-        "platform": {
-            "php": "8.0.99"
-        },
-        "sort-packages": true,
-        "allow-plugins": {
-            "dealerdirect/phpcodesniffer-composer-installer": true,
-            "composer/package-versions-deprecated": true,
-            "laminas/laminas-dependency-plugin": true
-        }
+        "source": "https://github.com/laminas/laminas-servicemanager",
+        "docs": "https://docs.laminas.dev/laminas-servicemanager/",
+        "rss": "https://github.com/laminas/laminas-servicemanager/releases.atom"
     },
     "require": {
-        "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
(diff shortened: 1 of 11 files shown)


src/ServiceManager.php
```diff
@@ -638,52 +630,6 @@ private function doCreate(string $resolvedName, ?array $options = null)
         return $object;
     }
 
-    /**
```
Ocramius: IMO worth checking proxy-manager-lts meanwhile: I didn't touch `ocramius/proxy-manager` in over 2 years, since the appearance of `proxy-manager-lts` removed all will of mine to maintain the original `proxy-manager`.

In other words, `proxy-manager-lts` makes my blood boil, but it is better to introduce it here than removing these critical tests and code here.

Ocramius (changes requested):

src/ServiceManager.php
Slamdunk: I feel like it shouldn't be `laminas/laminas-servicemanager` responsibility to provide this functionality *out-of-the-box*, and since the first release supporting lazy services 8 years ago, lazy loading is now *built-in* in pretty much every library I use nowadays.

But if you want me to replace `ocramius/proxy-manager` with `friendsofphp/proxy-manager-lts` instead of dropping it, ping me again and I'll do it.

Ocramius: Yes please

Slamdunk: Done.

composer.json
```diff
@@ -13,72 +13,78 @@
...
+        "laminas/laminas-code": "<4.10.0",
         "zendframework/zend-code": "<3.3.1",
         "zendframework/zend-servicemanager": "*"
     },
+    "provide": {
+        "psr/container-implementation": "^1.0"
+    },
     "suggest": {
-        "ocramius/proxy-manager": "ProxyManager ^2.1.1 to handle lazy initialization of services"
+        "friendsofphp/proxy-manager-lts": "ProxyManager ^2.1.1 to handle lazy initialization of services"
```
froschdesign, line 54: The documentation must also be updated: https://docs.laminas.dev/laminas-servicemanager/lazy-services/#setup

Slamdunk: Ops, thank you :+1:

Ocramius (approved):
Thanks @Slamdunk!

PR #189, opened 2023-05-02
Introduce `ConfigValidatorInterface` for minimum runtime validation of SM configurations

### LINKED ISSUE #131: Feature: ServiceManager configuration validation
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

Whenever we are working in projects with own plugin managers (like laminas-validator, etc.), we do probably ignore the configuration structure which is fetched from the container.

https://github.com/laminas/laminas-validator/blob/bdd503adc83d814a5c94e598ea0eb9fc7ca56339/src/ValidatorPluginManagerFactory.php#L49

---

So there will be some point in time where we have to validate the plugin managers configuration. To avoid having to do this manually, I'd prefer fetching a service from the container which is able to verify that the configuration is valid.

Such method/interface like I've created in the `laminas-cache` component would suffice.

https://github.com/laminas/laminas-cache/blob/130cc8db636d97b8287ff817515bb1c9ae0ef099/src/Service/StorageAdapterFactoryInterface.php#L45

---

WDYT? Feedback very welcome.

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes

### Description

Starting with v4.0.0, the `ServiceManager` itself will not validate anymore as it provides psalm array-shape type for configuration values.

As per existing projects, the `ServiceManager` is usually configured somehow like these examples:

```php
$smConfig = $configuration['service_manager'] ?? [];
$smConfig = new ServiceManagerConfig($smConfig);

$serviceManager = new ServiceManager();
$smConfig->configureServiceManager($serviceManager);
```

```php
$config = require __DIR__ . '/config.php';

$dependencies                       = $config['dependencies'];
$dependencies['services']['config'] = $config;

$serviceManager = new ServiceManager($dependencies);
```

The first example does use deprecated `Config` instance which just verified that pre-defined array keys are provided.
The second example does not verify anything and assumes the config to be correct.

Both cases do either need a psalm-baseline entry or a `psalm-suppress` once migrated to SM v4.0.0 (at least the second example already needs that as it uses the `ServiceManager#__construct` which is already properly typed.

Closes #131 

---

So every pluginmanager providing component (such as `laminas-cache`, `laminas-form`, etc. - I've listed some more in the [TSC meeting](https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2023--1-09-TSC-Minutes.md#psr-11-v2)) would also need either [a baseline entry or suppression](https://github.com/laminas/laminas-form/blob/e49893f4a3f1fbe3e70238b119366b6913cc83ef/src/FormElementManagerFactory.php#L51) due to the fact that there is no current way to verify that configuration matches expectations.

To avoid that, we could in fact use a runtime configuration validator which is provided by this PR.

----

I have not yet a clue on how to effectively use it, as I would actually prefer to only validate prior configuration caching so that not every request has to validate the same valid/invalid config over and over again. I've created this PR as a draft so that we can find a way of how to effectively use such a validator. In mezzio projects, having this as some kind of `PostProcessor` for `config-aggregator` would work but that won't work for i.e. [laminas-form](https://github.com/laminas/laminas-form/blob/e49893f4a3f1fbe3e70238b119366b6913cc83ef/src/FormElementManagerFactory.php#L51)

Would love to get some help here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "laminas/laminas-dependency-plugin": "^2.2",
         "mikey179/vfsstream": "^1.6.11",
         "phpbench/phpbench": "^1.2.9",
-        "phpunit/phpunit": "^10.0.17",
+        "phpunit/phpunit": "^10.1",
         "psalm/plugin-phpunit": "^0.18.4",
         "vimeo/psalm": "^5.8.0"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "bcc74076fd03d8f52e6d9320feeaf24e",
+    "content-hash": "744296c1951e44221fda0b58df6deaf8",
     "packages": [
         {
             "name": "laminas/laminas-stdlib",
@@ -2051,16 +2051,16 @@
         },
         {
             "name": "phpunit/php-code-coverage",
-            "version": "10.1.0",
+            "version": "10.1.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sebastianbergmann/php-code-coverage.git",
-                "reference": "fc4f5ee614fa82d50ecf9014b51af0a9561f3df8"
+                "reference": "884a0da7f9f46f28b2cb69134217fd810b793974"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sebastianbergmann/php-code-coverage/zipball/fc4f5ee614fa82d50ecf9014b51af0a9561f3df8",
-                "reference": "fc4f5ee614fa82d50ecf9014b51af0a9561f3df8",
+                "url": "https://api.github.com/repos/sebastianbergmann/php-code-coverage/zipball/884a0da7f9f46f28b2cb69134217fd810b793974",
+                "reference": "884a0da7f9f46f28b2cb69134217fd810b793974",
                 "shasum": ""
             },
             "require": {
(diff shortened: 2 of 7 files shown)


gsteel:
Could we wire up a service in the config provider here so that components that ship plugin managers can use that service to perform runtime validation in the plugin manager factory?

Say a service something like this:

```php

namespace Laminas\ServiceManager;

/** @psalm-import-type ServiceManagerConfiguration from ServiceManager */
final readonly class ConfigValidationService
{
    public function __construct(private ValidatorInterface $validator, private bool $enabled)
    {
    }

    /** @psalm-assert ServiceManagerConfiguration $config */
    public function validate(array $config): void
    {
        if (! $this->enabled) {
            return;
        }

        $this->validator->assertValidConfig($config);
    }
}
```

… where `$this->enabled` is equivalent to `$config['debug']` and in a plugin manager factory, something like:

```php
final class PluginManagerFactory
{
    public function __invoke(ContainerInterface $container): ValidatorPluginManager
    {
        $config = $container->get('config');
        $services = $config['validators'] ?? [];
        
        $validator = $container->get(ValidatorInterface::class);
        $validator->validate($services);
        
        return new PluginManager($services);
    }
}
```

kynx:
> I have not yet a clue on how to effectively use it, as I would actually prefer to only validate prior configuration caching so that not every request has to validate the same valid/invalid config over and over again.

Yeah. I'd argue that validating configuration is a job for whatever build tool you use, not a runtime check.

As a developer, if I'm changing the config it's going to be obvious pretty quickly if an `alias` I've introduced does not resolve. If it's set to something different in production that's a problem. But does throwing an exception when the configuration is first loaded - as opposed to when the invalid dependency is first used - really make my life any easier? The borked config should never have got so far.

 Maybe these could be made into reusable smoke tests instead?

PR #193, opened 2023-05-10
Mark `Config` and `ConfigInterface` as deprecated

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Marking the `ConfigInterface` and the `Config` object. They do not provide any useful additions besides the fact that it holds a configuration which could also get passed as an array. In most if not all use-cases, the config is passed as an array to the `ServiceManager` anyways and thus, having an array should suffice as well.


boesing:
Merging as phpunit issues are unrelated

fezfez:
Hello @boesing ,

Just a question, im getting this error with phpstant

```

PR #200, opened 2023-07-02
Replace scope to retrieve plugin manager from `$this` to `self`

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Some parts of the `CommonPluginManagerTrait` do still use `$this` for the static method `CommonPluginManagerTrait#getPluginManager`

### DIFF
--- a/src/Test/CommonPluginManagerTrait.php
+++ b/src/Test/CommonPluginManagerTrait.php
@@ -26,7 +26,7 @@ trait CommonPluginManagerTrait
 {
     public function testInstanceOfMatches(): void
     {
-        $manager    = $this->getPluginManager();
+        $manager    = self::getPluginManager();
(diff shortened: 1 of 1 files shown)


src/Test/CommonPluginManagerTrait.php
```diff
@@ -26,20 +26,20 @@ trait CommonPluginManagerTrait
 {
     public function testInstanceOfMatches(): void
     {
-        $manager    = $this->getPluginManager();
+        $manager    = self::getPluginManager();
```
Ocramius, line 29: Could we perhaps drop the `CommonPluginManagerTrait`? :|

Ocramius (approved):
LGTM meanwhile :+1:

PR #204, opened 2023-08-27
Introduce plugin manager static analysis file

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Test is currently failing, I will debug this file with psalm to provide a proper fix to psalm itself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
     "homepage": "https://laminas.dev",
     "support": {
         "issues": "https://github.com/laminas/laminas-servicemanager/issues",
-        "forum": "https://discourse.laminas.dev",
         "chat": "https://laminas.dev/chat",
         "forum": "https://discourse.laminas.dev"
     },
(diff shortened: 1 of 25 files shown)


boesing:
Psalm fix is available in https://github.com/vimeo/psalm/pull/10157

src/AbstractSingleInstancePluginManager.php
```diff
@@ -26,14 +26,6 @@ abstract class AbstractSingleInstancePluginManager extends AbstractPluginManager
      */
     protected string $instanceOf;
 
-    /**
```
boesing, line 29: Removing type-narrowing here as it actually breaks psalms conditional detection.
I do not really have time to improve this right now and thus removing this for now.
If someone still manages to return invalid stuff, that would be on purpose and thus we should not really care about it.

composer.json
```diff
@@ -53,7 +52,7 @@
         "phpunit/phpunit": "^10.1",
         "psalm/plugin-phpunit": "^0.18.4",
         "symfony/console": "^6.0",
-        "vimeo/psalm": "^5.10"
+        "vimeo/psalm": "5.x-dev as 5.16"
```
boesing: - [x] Update to stable version once released

PR #207, opened 2023-10-10
Add support for PHP 8.3

glo71317:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,13 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
-         colors="true"
-         cacheDirectory=".phpunit.cache">
-  <coverage>
-    <include>
-      <directory suffix=".php">./src</directory>
-    </include>
-  </coverage>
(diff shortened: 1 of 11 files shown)


froschdesign:
@glo71317 
Please use the following branch: [3.22.x](https://github.com/laminas/laminas-servicemanager/tree/3.22.x)

@boesing 
Will this work or is there something else that needs to be considered or is there another plan for how to proceed?

boesing:
Should work, targeting latest 3.x branch, create new minor, merge-up to v4, create new RC.
I can take over here when I find some time, maybe this weekend.

We also need `ignore_php_platform_requirements` in the `.laminas-ci.json` where we add 8.3 to avoid transitive dependency incompatibilities. We do want to verify if this package is compatible, we do not (yet) care about transitive dependencies until these packages have issues with PHP 8.3 and do have impact on this component

glo71317:
@froschdesign i have created new pr -https://github.com/laminas/laminas-servicemanager/pull/209 against branch 3.22.x

boesing:
`README` linting is fixed in #210

phpunit.xml.dist
```diff
@@ -18,4 +11,9 @@
       <group>disable</group>
     </exclude>
   </groups>
+  <source>
```
boesing, line 14: @laminas/technical-steering-committee any ideas (besides "do not run tests on latest") to somehow have tests not failing on "latest" when phpunit changes config? No clue why sebastian is dropping still compatible (but deprectated) configuration flags entirely even tho they still work...
So config was valid, even for the "latest" version but as per the schema, it was not.
Should I ask in phpunit why there are config flags dropped from the schema instead of marking them deprecated? Does XSD even support the concept of "deprecated"?

boesing:
Thanks, @glo71317!

PR #213, opened 2023-10-14
Narrow array type for `services` in ServiceManager configuration

### LINKED ISSUE #212: Array shape of ServiceManager config contains unspecific array
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 3.22.0

#### Summary

Not really a bug but more of a QA issue.

The `ServiceManager`s `ServiceManagerConfiguration` array shape contains a `services` key with an unspecific `array` notation. That trips some static analyzers.

#### Current behavior

`ServiceManagerConfiguration` contains the definition: `services?: array<string,object|array>,` which includes a generic `array`.

#### How to reproduce

Use the `ServiceManagerConfiguration` configuration anywhere and run PHPStan (I tried 1.10) which emits `PHPDoc tag @var for variable $serviceManagerConfig has no value type specified in iterable type array.`

#### Expected behavior

PHPStan should not throw an error. This is achieved by using a more "specific" array definition: `array<array-key, mixed>`.

rieschl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR adds a more specific array shape for the `services` key in `ServiceManagerConfiguration`.

fixes #212

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -89,7 +89,7 @@ use const E_USER_DEPRECATED;
  *     initializers?: InitializersConfiguration,
  *     invokables?: array<string,string>,
  *     lazy_services?: LazyServicesConfiguration,
- *     services?: array<string,object|array>,
+ *     services?: array<string,object|array<mixed>>,
  *     shared?:array<string,bool>,
  *     shared_by_default?:bool,
  *     ...


src/ServiceManager.php
```diff
@@ -89,7 +89,7 @@
  *     initializers?: InitializersConfiguration,
  *     invokables?: array<string,string>,
  *     lazy_services?: LazyServicesConfiguration,
- *     services?: array<string,object|array>,
+ *     services?: array<string,object|array<array-key,mixed>>,
```
boesing: As requested in #212, please use `array<mixed>` instead.

boesing (changes requested):

src/ServiceManager.php
boesing: @laminas/technical-steering-committee should we enforce this kind of `array` notation via our codestandard and if so, are there existing rules for that? That would prevent us from having these kind of "issues" with incompatibilities between psalm/phpstan in the future.

boesing (approved):
LGTM, ~will see when I find some time to prepare a release~

nvm, is just a patch. going to release now.

rieschl:
> LGTM, ~will see when I find some time to prepare a release~
> 
> nvm, is just a patch. going to release now.

thanks!

PR #217, opened 2023-10-29
Update composer.json allow psr-container 2.0

markusramsak:


Ocramius:
Not possible in 3.x, since this is a breaking change.

hectorprats:
What is breaking in laminas-servicemanager with this change?

froschdesign:
@hectorprats 
Check older pull requests to this topic, for example #175 and #147.

PR #230, opened 2024-03-13
Make `ServiceLocatorInterface#get()` generic

### LINKED ISSUE #228: Override/improve ContainerInterface::get return type hints in ServiceLocatorInterface (generics, class-string)
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Currently, ServiceLocatorInterface extends ContainerInterface such that only the `build` method is added.
The `get` method is inherited by ContainerInterface and not modified.
ServiceManager extends ServiceLocatorInterface and uses `{@inheritDoc}` for `get` so it inherits whatever PHPDoc ContainerInterface and ServiceLocatorInterface define (FTR `@inheritDoc` can be dropped safely I think).

I suggest to override the `get` in ServiceLocatorInterface so that it behaves the same as `build` for Psalm and PHPStan when the input is a class-string. Should be as simple as adding:
```
    /**
     * Finds an entry of the container by its identifier and returns it.
     *
     * @template T of object
     * @param string|class-string<T> $id Identifier of the entry to look for.
     * @return mixed Entry.
     * @psalm-return ($id is class-string<T> ? T : mixed)
     * @throws ContainerExceptionInterface Error while retrieving the entry.
     * @throws NotFoundExceptionInterface  No entry was found for **this** identifier.
     */
    public function get(string $id);
```

The only drawback is that ServiceLocatorInterface then becomes prone to ContainerInterface method signature changes. The last change to the `ContainerInterface::get` was made in psr/container version 1.1.0, so version 1.0.0 would be incompatible but all other version up to / including 2.0.2 would be compatible.

InvisibleSmiley:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Fix #228: Override/improve ContainerInterface::get return type hints in ServiceLocatorInterface (add generics / let class-string input affect return type inference)

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -211,6 +211,11 @@
       <code><![CDATA[MixedPluginManager]]></code>
     </UnusedClass>
   </file>
+  <file src="test/StaticAnalysis/ServiceLocatorInterfaceConsumer.php">
+    <UnusedClass>
+      <code><![CDATA[ServiceLocatorInterfaceConsumer]]></code>
+    </UnusedClass>
+  </file>
   <file src="test/StaticAnalysis/ServiceManagerConfiguration.php">
     <UnusedClass>
       <code><![CDATA[ServiceManagerConfiguration]]></code>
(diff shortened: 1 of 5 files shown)


boesing (changes requested):
LGTM. 
Could you please add a static analysis test in `test/StaticAnalysis` to verify that psalm is able to infer the return value as expected?

Ocramius (approved):
Looking good! 🚢

PR #238, opened 2024-06-09
reflection factory resolves array $options

dappvibe:
To pass $options on `$container->build($name, $options);`


Ocramius (changes requested):
I'd first start from a test explaining why this change is necessary :-)

PR #240, opened 2024-10-07
PHP8.4 Support Added

cod43156:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no


boesing:
FYI you can safely drop the psalm plugin as it does not work anymore.

gsteel:
This patch is no longer necessary @cod43156 

We added PHP 8.4 support to the 3.x series in #243 and released 3.23.0 - Support in the 4.x series will be merged via #246 and released as 4.3.0 in due course.

Thanks for the help!

PR #243, opened 2024-10-18
Add PHP 8.4 Support to the 3.x series

gsteel:
### Description

Adds 8.4 support to the 3.x series

- Ignores platfrom reqs on 8.4
- Fixes a bunch of docs linting problems
- Bumps stdlib to an 8.4 compatible release
- Bumps all dev deps
- Baselines/fixes a number of new Psalm issues - most of this is around the types being moved from ConfigInterface to ServiceManager, and changes to Psalm's reporting of certain issues.

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,5 +1,6 @@
 {
     "ignore_php_platform_requirements": {
-        "8.3": true
+        "8.3": true,
+        "8.4": true
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -21,21 +21,21 @@
         "rss": "https://github.com/laminas/laminas-servicemanager/releases.atom"
     },
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
-        "laminas/laminas-stdlib": "^3.17",
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
+        "laminas/laminas-stdlib": "^3.19",
         "psr/container": "^1.0"
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11.99.5",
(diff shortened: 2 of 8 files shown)


.laminas-ci.json
```diff
@@ -1,5 +1,6 @@
 {
     "ignore_php_platform_requirements": {
-        "8.3": true
+        "8.3": true,
```
froschdesign, line 3: Version 8.3 is still needed here?

gsteel: Yes: https://github.com/laminas/laminas-servicemanager/actions/runs/11406081441

alexmerlin (approved):
LGTM!

composer.json
```diff
@@ -21,21 +21,21 @@
...
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
+        "laminas/laminas-stdlib": "^3.19",
         "psr/container": "^1.0"
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11.99.5",
-        "friendsofphp/proxy-manager-lts": "^1.0.14",
-        "laminas/laminas-code": "^4.10.0",
+        "friendsofphp/proxy-manager-lts": "^1.0.18",
+        "laminas/laminas-code": "^4.14.0",
```
froschdesign, line 31: Should a version that supports 8.4 also be added here? ([4.15.0](https://github.com/laminas/laminas-code/releases/tag/4.15.0))

gsteel: CI with 8.4 on lowest doesn't have a problem with it and as a dev dep, it can be bumped at any time, so I don't think it is _necessary_ but happy to bump again if you do 👍

froschdesign: If it is not necessary, then it can stay that way.

froschdesign (approved):
LGTM 👍🏻

PR #244, opened 2024-10-21
Add mixed as array shape

grizzm0:
Psalm treats `array` as `array<mixed>` but PHPStan on level 6 and above requires a shape on the array and gives error when importing `ServiceManagerConfiguration`.


samsonasik (approved):

gsteel (approved):
Thanks @grizzm0 LGTM 👍

PR #249, opened 2024-12-11
fix: Fix LazyServiceFactory to handle exception in constructor

idsulik:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
This PR fixes a bug where exceptions thrown during lazy service initialization would leave the proxy in an invalid state. The fix ensures that the original initializer is restored when an exception occurs, allowing subsequent initialization attempts to work correctly.
Added tests to verify the behavior.

p.s. thanks to @vdmorozov, he found the issue

### DIFF
--- a/src/Proxy/LazyServiceFactory.php
+++ b/src/Proxy/LazyServiceFactory.php
@@ -10,6 +10,7 @@ use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
 use ProxyManager\Proxy\VirtualProxyInterface;
 use Psr\Container\ContainerInterface;
+use Throwable;
 
 use function sprintf;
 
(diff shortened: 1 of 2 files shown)


idsulik:
if it's possible, please create a patch release for the 3 version with this fix

src/Proxy/LazyServiceFactory.php
```diff
@@ -42,8 +43,14 @@ public function __invoke(
...
         if (isset($this->servicesMap[$name])) {
             $initializer = static function (&$wrappedInstance, LazyLoadingInterface $proxy) use ($callback): bool {
+                $initializer = $proxy->getProxyInitializer();
                 $proxy->setProxyInitializer(null);
-                $wrappedInstance = $callback();
+                try {
+                    $wrappedInstance = $callback();
+                } catch (Throwable $e) {
+                    $proxy->setProxyInitializer($initializer);
+                    throw $e;
```
vdmorozov, line 52: What if we just swap these two rows: `$proxy->setProxyInitializer(null)` and `$wrappedInstance = $callback()`?

It'd look like this:
```php
                $wrappedInstance = $callback();
                $proxy->setProxyInitializer(null);
```

idsulik:
@vdmorozov will create a separate PR

vdmorozov:
Here: https://github.com/laminas/laminas-servicemanager/pull/250

PR #250, opened 2024-12-11
fix: LazyServiceFactory does not reset initializer in proxy class when building dependency fails

vdmorozov:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR fixes a bug where an exception thrown during lazy service initialization would leave the proxy in an invalid state. The fix ensures that the original initializer isn't reset when an exception occurs, allowing subsequent initialization attempts to work correctly.

Added tests to verify the behavior.


vdmorozov:
If possible, please create a patch release for the 3-rd version with this fix

PR #251, opened 2024-12-13
Remove duplicate delgators for all services

settermjd:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

[`ServiceManager::mergeDelegators()` documentation](https://github.com/laminas/laminas-servicemanager/blob/4.4.x/src/ServiceManager.php#L697-L699) says that it merges delegators avoiding multiple same delegators for the same service. However, from experimenting with it, duplicates **are** allowed. So, I've created this small patch to ensure that duplicates are removed, along with a covering test for it to verify the change.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -705,8 +705,7 @@ class ServiceManager implements ServiceLocatorInterface
     {
         foreach ($config as $key => $delegators) {
             if (! array_key_exists($key, $this->delegators)) {
-                $this->delegators[$key] = $delegators;
-                continue;
+                $this->delegators[$key] = [];
             }
 
             foreach ($delegators as $delegator) {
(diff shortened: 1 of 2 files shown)


alexmerlin (approved):
Checked the content of `$this->delegators`, before the fix:
```text
array(1) {
  ["DateTime"]=>
  array(2) {
    [0]=>
    string(54) "Laminas\ContainerConfigTest\TestAsset\DelegatorFactory"
    [1]=>
    string(54) "Laminas\ContainerConfigTest\TestAsset\DelegatorFactory"
  }
}
```
and after the fix:
```text
array(1) {
  ["DateTime"]=>
  array(1) {
    [0]=>
    string(54) "Laminas\ContainerConfigTest\TestAsset\DelegatorFactory"
  }
}
```
This fixes the duplicate delegator issue.

PR #253, opened 2025-01-26
Use Attribute instead of annotation in Plugin Manager test Trait

gsteel:
In PHPUnit 11 doc-block comments are deprecated. This patch fixes tests in dependents that have already upgraded to v11


gsteel:
Needs to rebased after #252 is resolved

PR #255, opened 2025-01-26
Upgrade Psalm to v6

gsteel:
- Removes some unnecessary Psalm plugins
- General SA QA

<del>As an alternative to changes made in #252 , try upgrading to Psalm v6 to see if callable intersections in parameterized class-strings work now.</del>

Also,

- Closes #254 
- Closes #253 
- Closes #252

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -26,7 +26,7 @@
             <group>disable</group>
         </exclude>
     </groups>
-    <source>
+    <source restrictDeprecations="true">
         <include>
             <directory suffix=".php">./src</directory>
         </include>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,19 +1,34 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd" colors="true" cacheDirectory=".phpunit.cache">
-  <coverage/>
-  <testsuites>
-    <testsuite name="laminas-servicemanager Test Suite">
-      <directory>./test/</directory>
-    </testsuite>
-  </testsuites>
-  <groups>
-    <exclude>
-      <group>disable</group>
-    </exclude>
-  </groups>
-  <source>
-    <include>
-      <directory suffix=".php">./src</directory>
-    </include>
-  </source>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+         colors="true"
+         cacheDirectory=".phpunit.cache"
+         displayDetailsOnPhpunitDeprecations="true"
+         displayDetailsOnTestsThatTriggerWarnings="true"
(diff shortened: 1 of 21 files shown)


src/AbstractPluginManager.php
```diff
@@ -255,9 +255,7 @@ public function setInvokableClass(string $name, string|null $class = null): void
      *
      * @deprecated Please use {@see AbstractPluginManager::configure()} instead.
      *
-     * @phpcs:disable Generic.Files.LineLength.TooLong
-     * @param class-string<FactoryInterface>|class-string<object&FactoryCallable>|FactoryCallable|FactoryInterface $factory
-     * @phpcs:enable Generic.Files.LineLength.TooLong
+     * @param class-string<FactoryInterface>|FactoryCallable|FactoryInterface $factory
```
gsteel, line 258: Using intersections with callable in a class string parameter is not supported, i.e. `class-string<object&callable>` causes an `InvalidDocblock` issue in Psalm

src/AbstractPluginManager.php
```diff
@@ -301,9 +299,7 @@ public function addAbstractFactory(string|AbstractFactoryInterface $factory): vo
      *
      * @param string $name Service name
      * @param string|callable|DelegatorFactoryInterface $factory Delegator factory to assign.
-     * @phpcs:disable Generic.Files.LineLength.TooLong
-     * @psalm-param class-string<DelegatorFactoryInterface>|class-string<object&DelegatorCallable>|DelegatorCallable $factory
-     * @phpcs:enable Generic.Files.LineLength.TooLong
+     * @psalm-param class-string<DelegatorFactoryInterface>|DelegatorCallable $factory
```
gsteel, line 302: Using intersections with callable in a class string parameter is not supported, i.e. `class-string<object&callable>` causes an `InvalidDocblock` issue in Psalm

phpunit.xml.dist
```diff
@@ -1,19 +1,34 @@
...
+        <testsuite name="laminas-servicemanager Test Suite">
+            <directory>./test/</directory>
+        </testsuite>
+    </testsuites>
+    <groups>
+        <exclude>
+            <group>disable</group>
+        </exclude>
+    </groups>
+    <source restrictWarnings="true" restrictNotices="true" restrictDeprecations="true">
```
gsteel: On deps = lowest, we end up with symfony/filesystem at <7.2 where they fixed implicit null deprecations on 8.4 - we can't restrict this transitive dep to ^7.2 because a) it would be weird, and b) it's 8.2+, so I'm ignoring deprecations triggered in 3rd party code instead

src/Test/CommonPluginManagerTrait.php
```diff
@@ -45,16 +46,15 @@ public function testLoadingInvalidElementRaisesException(): void
...
-     */
+    #[DataProvider('aliasProvider')]
     public function testPluginAliasesResolve(string $alias, string $expected): void
     {
         $this->assertInstanceOf($expected, self::getPluginManager()->get($alias), "Alias '$alias' does not resolve'");
     }
 
     /**
      * @return list<array{string,string}>
+     * @psalm-suppress PossiblyUnusedMethod
```
Ocramius, line 57: BTW, I fixed these by keeping (for now) both the attribute and thr annotation

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #266, opened 2025-10-07
"psr/container": "^1.1 || ^2.0"

v2catch:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
         "laminas/laminas-stdlib": "^3.17",
-        "psr/container": "^1.0"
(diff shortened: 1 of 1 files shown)


arhimede:
the psr container 1 OR 2 is implemented only on version 4 of service manager

froschdesign:
Adding the version 2 of the PSR container means also to update the related classes in this package but this ends in a BC break. Use version 4 of laminas-servicemanager if you want to use version 2 of the PSR-11 container interface.

See: https://docs.laminas.dev/laminas-servicemanager/v4/psr-11/

PR #268, opened 2025-10-14
Remove support for PHP 8.1, Add support for PHP 8.5

gsteel:
Bumps dev deps and fixes a bunch of minor SA issues.

Can't upgrade PHPUnit here without first upgrading container tests to support 11.x

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,6 +1,6 @@
 {
     "ignore_php_platform_requirements": {
-        "8.3": true,
-        "8.4": true
+        "8.4": true,
+        "8.5": true
     }
 }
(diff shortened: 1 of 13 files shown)


.laminas-ci.json
```diff
@@ -1,6 +1,7 @@
 {
     "ignore_php_platform_requirements": {
         "8.3": true,
-        "8.4": true
+        "8.4": true,
```
froschdesign, line 3: Unfortunately, I do not have an overview here, but do the dependencies still need to be ignored for all these versions?

gsteel: Yes - `laminas-container-config-test` is the culprit here. Might be OK for 8.3 though - hang on and I'll try

froschdesign (approved):

PR #270, opened 2025-10-14
Update PHPunit to 11

Xerkus:
Baselined new false positive psalm error with phpunit plugin not refining string to non-empty-string with `assertNotEmpty()`.


gsteel (approved):
Thanks @Xerkus

PR #271, opened 2025-10-20
doc: add laminas-servicemanager-migration link on v3-to-v4 migration doc

samsonasik:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

@Xerkus there is documentation for `laminas-servicemanager-migration`, this PR add link to the v3-to-v4 doc.


docs/book/v4/migration/v3-to-v4.md
```diff
@@ -1,5 +1,3 @@
 # Migration from Version 3 to 4
 
-INFO: **Work in progress**
-We are currently working on the migration guide.
-We will update this page as soon as possible!
+The ServiceManager 4 introduce new interfaces and updated the factory class structure. We have [Rector Rule](https://github.com/laminas/laminas-servicemanager-migration) migrations that cover that.
```
froschdesign: It does not look as if all the changes have been covered: https://github.com/laminas/laminas-servicemanager/releases/tag/4.0.0

froschdesign: Here is an example of a good migration guide: https://docs.laminas.dev/laminas-filter/v3/migration/v2-to-v3/

samsonasik: I've updated with list of changes https://github.com/laminas/laminas-servicemanager/pull/271/commits/18be48c256dec21f824a7a1171aaa27161e4b453

docs/book/v4/migration/v3-to-v4.md
```diff
@@ -1,5 +1,47 @@
 # Migration from Version 3 to 4
 
-INFO: **Work in progress**
-We are currently working on the migration guide.
-We will update this page as soon as possible!
```
Xerkus, line 5: Since this is a partial migration guide I suggest we keep this disclaimer until full migration guide is available

samsonasik: sure, I rolled back disclaimer https://github.com/laminas/laminas-servicemanager/pull/271/commits/30eaf168244d379ffbe96684dca8899b096aeb9f

docs/book/v4/migration/v3-to-v4.md
```diff
@@ -3,3 +3,49 @@
...
+use Psr\Container\ContainerInterface;
+
+class ServiceFactory
+{
+    public function __invoke(ContainerInterface $container)
+    {
+
+    }
+}
+```
```
Xerkus, line 51: This is not really an equivalent change. "After" here should be implementing new interface.

For factories with no interface before/after needs to be like this after example with only `s/Interop/Psr/`

samsonasik: This is not necessarily needs implements new interface, factory can just class with __invoke, see

- https://github.com/laminas/laminas-servicemanager-migration/issues/34

Xerkus: It was not necessary to implement old interface either.

Xerkus: support both versions if factory is the only thing depending on SM -> remove interface if used, swap Interop to Psr.

keeping interface -> replace interface

had no interface -> swap Interop to Psr if not done already. They were aliases.

samsonasik: The rule already cover if the `$requestedName` or `$options` is not used, so just use as is. The old interface enforce `$requestedName` and `$options`.

samsonasik: see that it skipped if $options or $requestedName is used

https://github.com/laminas/laminas-servicemanager-migration/blob/674d1d63a56b627869f97a61b255df0cc58386de/test/Rector/Class_/ImplementsFactoryInterfaceToPsrFactoryRector/Fixture/skip_options_used.php.inc#L10-L16

https://github.com/laminas/laminas-servicemanager-migration/blob/674d1d63a56b627869f97a61b255df0cc58386de/test/Rector/Class_/ImplementsFactoryInterfaceToPsrFactoryRector/Fixture/skip_requested_name_and_options_used.php.inc#L10-L16

Xerkus: Yes,  but this is not the rector tool documentation. This needs to cover service manager changes not rector rule preferences.

samsonasik: I see, what before vs after needs to be used here?

Xerkus: interface -> interface
not interface interop -> no interface psr

And mention of no interface psr factory supporting both versions

Xerkus: There is also removed legacy interface -> interface

samsonasik:
feel free to create alternative PRs with more complete list guys :)

PR #274, opened 2026-01-11
Can a service be built?

gsteel:
When you have a service registered under `services` without a backing factory, `has` will return true, and `get` will fetch it, but `build` will throw an exception.

This commit adds `ServiceLocatorInterface::canBuild($name)` so that users can determine whether a call to `build` will be successful or not.

BTW, the main reason I want to add this is for plugin managers. Often, validators, filters, inputs etc are not shared, and calling `build(name, options)` is typical behaviour because `get` no longer accepts options, and these types often require options arrays.

For user-defined _services_, we can't know if it is "safe" to build a service (i.e. the user registered a factory as opposed to dropping an instance into 'services').

### DIFF
--- a/src/AbstractPluginManager.php
+++ b/src/AbstractPluginManager.php
@@ -191,7 +191,7 @@ abstract class AbstractPluginManager implements PluginManagerInterface
                 throw new Exception\ServiceNotFoundException(sprintf(
                     'A plugin by the name "%s" was not found in the plugin manager %s',
                     $id,
-                    static::class
+                    static::class,
                 ));
             }
 
@@ -209,6 +209,11 @@ abstract class AbstractPluginManager implements PluginManagerInterface
     public function has(string $id): bool
     {
         return $this->plugins->has($id);
+    }
+
+    public function canBuild(string $name): bool
+    {
+        return $this->plugins->canBuild($name);
     }
 
     /**
--- a/src/ServiceLocatorInterface.php
+++ b/src/ServiceLocatorInterface.php
@@ -39,4 +39,12 @@ interface ServiceLocatorInterface extends ContainerInterface
      * @throws NotFoundExceptionInterface No entry was found for **this** identifier.
      */
     public function get(string $id);
+
+    /**
+     * Determines whether this container is capable of building a new instance of the given service name
+     *
+     * This method is useful for differentiating between something that has been registered as a service,
+     * without a backing factory and something that can be constructed via a factory or an abstract factory.
+     */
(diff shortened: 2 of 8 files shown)


Ocramius:
Not 100% sure we need this: could be useful for optional services, but anything calling `has()` as a failsafe should probably make things fail.

gsteel:
@Ocramius 

> Not 100% sure we need this: could be useful for optional services, but anything calling `has()` as a failsafe should probably make things fail.

The point here is that anything in `['services' => ['foo' => new Instance()]]` is groovy for `get` and `has`, but if you _know_ you have a `ServiceLocatorInterface` such as a plugin manager, all you can do is try/catch if you need to `build` in order to pass in construction time options.

I agree that its use-cases are limited, but `services` does exist, so it's a useful tool when `build`-ing is what you need to do

Edit… 

Failing during `build` (as it does without this patch) _is_ an option, but it takes away the ability to present an exception such as "Your plugin cannot be built" rather than the chain of errors returned directly by ServiceMananger, so I guess the real use-case of this is so that we can provide a clearer exception to users where necessary

src/AbstractPluginManager.php
```diff
@@ -211,6 +211,11 @@ public function has(string $id): bool
         return $this->plugins->has($id);
     }
 
+    public function canBuild(string $name): bool
+    {
+        return $this->plugins->canBuild($name);
+    }
```
Ocramius, line 217: Let's have this on its own interface, even if it's doing just that?

src/ServiceLocatorInterface.php
```diff
@@ -39,4 +39,12 @@ public function build(string $name, ?array $options = null): mixed;
...
      */
     public function get(string $id);
+
+    /**
+     * Determines whether this container is capable of building a new instance of the given service name
+     *
+     * This method is useful for differentiating between something that has been registered as a service,
+     * without a backing factory and something that can be constructed via a factory or an abstract factory.
+     */
+    public function canBuild(string $name): bool;
```
Ocramius, line 49: I wouldn't expand this existing interface: expanding the interface is also a major BC break

Ocramius (changes requested):

gsteel:
Thanks for the review here @Ocramius 

Apart from the obvious BC breaks, I'm closing anyway… I needed this in refactoring of `input-filter`, but I've decided to instead go with 

```php
try {
    $pluginManager->build($thing, $options);
} catch (AppropriateException) {
    $pluginManager->get($thing);
}
```

… or similar

PR #280, opened 2026-04-14
Do not remove delegators on service creation

MartinKrizan:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

I have encountered bug in application which occurs when two requests triggers creation of same service at same time(app is running on swoole so ServiceManager is shared across requests).  In this case both requests call $serviceManager->get(Service), which leads to two doCreate calls. First call removes delegators and then second one creates service without them. Problem is that the result from the second call(without delegators) is stored in ServiceManager->services.

### DIFF
--- a/src/ServiceManager.php
+++ b/src/ServiceManager.php
@@ -601,9 +601,9 @@ class ServiceManager implements ServiceLocatorInterface
         $resolvedDelegators = [];
         foreach ($this->delegators[$name] as $index => $delegatorFactory) {
             /** @psalm-suppress ArgumentTypeCoercion https://github.com/vimeo/psalm/issues/9680 */
-            $delegatorFactory                = $this->resolveDelegatorFactory($delegatorFactory);
-            $this->delegators[$name][$index] = $delegatorFactory;
-            $creationCallback                =
+            $delegatorFactory           = $this->resolveDelegatorFactory($delegatorFactory);
+            $resolvedDelegators[$index] = $delegatorFactory;
+            $creationCallback           =
                 static fn(): mixed => $delegatorFactory($initialCreationContext, $name, $creationCallback, $options);
         }
 
--- a/test/ServiceManagerTest.php
+++ b/test/ServiceManagerTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace LaminasTest\ServiceManager;
 
 use DateTime;
+use Laminas\ContainerConfigTest\TestAsset\Delegator;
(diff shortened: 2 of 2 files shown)


mairo744:
We encountered this issue after upgrading to service manager v4 in our swoole based projects (mezzio-swoole). It was quite difficult to investigate because it occurred only rarely, and after a worker refresh (e.g., after reaching max requests), the issue disappeared.

We deployed the following fix to production, and so far it seems to have resolved the problem — we haven’t observed any occurrences since.

It appears the issue might have been introduced in this commit:
https://github.com/laminas/laminas-servicemanager/commit/74604b409b85b4cf9cb77f16b91229098087f299#diff-c88e8922673ece50d04aef21f4dc95b87bd465868eb073480467f87e626e3b0aR550

image


Maybe @boesing remembers the reasoning behind this change, although it dates back to April 20, 2023.

If anyone has time to take a look at this, we’d really appreciate it. We’re also happy to assist further if something is missing or needs clarification.

boesing:
I will try to wrap my head around this this weekend. I do not really remember. Afair, this was also the case previously and/or was not covered by existing tests ans got added as a regression.

boesing:
Hey guys,

tbh I do not really remember why this got removed. If this PR is restoring a behavior which was there in v3, I do not see why we should not restore this PLUS having a test for it to avoid removal in future refactorings.

Since the existing tests are still passing, I'd say, lets merge it - yet I do not want to make any decision here as I havent participated in the most recent changes regarding laminas.

To me, it makes sense having the services passing the delegator on multiple calls - yet I guess the code I wrote was expecting services to be cached due to shared=true. Am I correct, that this behavior only affects non-shared services?

Maybe @gsteel can have a quick look at this and merge it to release a fixed version.

MartinKrizan:
Hi, 
thanks for the review.

It also affects shared services, but only in async environment when ServiceManager is shared across parallel tasks.  Here is a test that reproduces the behavior with shared service on swoole https://github.com/MartinKrizan/laminas-servicemanager/commit/b2a0ce0151b9cea57080513c37b193c4b6b807a1

mairo744:
Just friendly reminder
The fix has been running stable in production on our side

froschdesign (approved):

froschdesign:
@MartinKrizan
Thank you for your time and this contribution! 👍

@boesing @mairo744 
Thanks for you help! 👍🏻

